PR #1, opened 2022-02-24
Docs

lovromazgon:
This adds a readme and improves the docs on public structs.

It also relaxes two requirements when implementing plugins:
- Implementing the `Source.Ack` method is not required anymore, the SDK will ignore the `ErrUnimplemented` error.
- Implementing the `Destination.Flush` method is only required if `WriteAsync` is implemented, otherwise the SDK will ignore the `ErrUnimplemented` error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,52 @@
 # Conduit Connector Plugin SDK
 
-Readme coming soon. :construction:
+[![License](https://img.shields.io/badge/license-Apache%202-blue)](https://github.com/ConduitIO/connector-plugin-sdk/blob/main/LICENSE.md)
+[![Build](https://github.com/ConduitIO/connector-plugin-sdk/actions/workflows/build.yml/badge.svg)](https://github.com/ConduitIO/connector-plugin-sdk/actions/workflows/build.yml)
+[![Go Report Card](https://goreportcard.com/badge/github.com/conduitio/connector-plugin-sdk)](https://goreportcard.com/report/github.com/conduitio/connector-plugin-sdk)
+[![Go Reference](https://pkg.go.dev/badge/github.com/conduitio/connector-plugin-sdk.svg)](https://pkg.go.dev/github.com/conduitio/connector-plugin-sdk)
+
+This repository contains the Go software development kit for implementing a connector plugin for
+[Conduit](https://github.com/conduitio/conduit). If you want to implement a connector plugin in another language please
+have a look at the [connector plugin protocol](https://github.com/conduitio/connector-plugin).
+
+## Quickstart
+
+Create a new folder and initialize a fresh go module:
+```
+go mod init example.com/conduit-plugin-demo
+```
+
+Add the connector plugin SDK dependency:
+
+```
+go get github.com/conduitio/connector-plugin-sdk
+```
+
+You need to create two structs, one that implements `sdk.Source` and another that implements `sdk.Destination`. Apart
+from that, you need to create three constructor functions where each returns a `sdk.Source`, `sdk.Destination` and
+`sdk.Specification` respectively.
+
+The plugin also needs an entrypoint that only calls `sdk.Serve`.
+
+```go
+package main
+
+import (
+	demo "example.com/conduit-plugin-demo"
+	sdk "github.com/conduitio/connector-plugin-sdk"
+)
+
+func main() {
+	sdk.Serve(
+		demo.Specification,  // func Specification() sdk.Specification { ... }
+		demo.NewSource,      // func NewSource() sdk.Source { ... }
+		demo.NewDestination, // func NewDestination() sdk.Destination { ... }
+	)
+}
+```
+
+## Examples
+
+For examples of simple plugins you can look at existing plugins like
+[conduit-plugin-generator](https://github.com/ConduitIO/conduit-plugin-generator) or
+[conduit-plugin-file](https://github.com/ConduitIO/conduit-plugin-file).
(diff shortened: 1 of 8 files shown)


destination.go
```diff
@@ -214,7 +226,10 @@ func (a *destinationPluginAdapter) waitForAcks(ctx context.Context) error {
 func (a *destinationPluginAdapter) Stop(ctx context.Context, req cpluginv1.DestinationStopRequest) (cpluginv1.DestinationStopResponse, error) {
 	// flush cached records
 	err := a.impl.Flush(ctx)
-	if err != nil {
+	if err != nil &&
+		// only propagate error if we are sure the plugin is writing records
+		// asynchronously or if it's some other error than ErrUnimplemented
+		(a.isAsync || !errors.Is(err, ErrUnimplemented)) {
```
lovromazgon, line 232: Relevant code snippet for making `Destination.Flush` optional for plugins that synchronously write records.

source.go
```diff
@@ -193,7 +198,8 @@ func (a *sourcePluginAdapter) runAck(ctx context.Context, stream cpluginv1.Sourc
 				return fmt.Errorf("ack stream error: %w", err)
 			}
 			err = a.impl.Ack(ctx, req.AckPosition)
-			if err != nil {
+			// implementing Ack is optional
+			if err != nil && !errors.Is(err, ErrUnimplemented) {
```
lovromazgon, line 202: Relevant code snippet for making `Source.Ack` optional.

README.md
```diff
@@ -1,3 +1,77 @@
...
+go mod init example.com/conduit-plugin-demo
+```
+
+Add the connector plugin SDK dependency:
+
+```
+go get github.com/conduitio/connector-plugin-sdk
+```
+
+You need to create two structs, one that implements `sdk.Source` and another that implements `sdk.Destination`. Apart
```
maha-hajja, line 25: one of these could be nil, maybe we should talk more about how one of these could be optional and how to only create a source or a destination connector

README.md
```diff
@@ -1,3 +1,77 @@
...
+```
+
+Now you can build the standalone plugin:
+
+```
+go build path/to/main.go
+```
+
+You will get a compiled binary which Conduit can use as a plugin. To run your plugin as part of a Conduit pipeline you
+can create a connector and specify the path to the compiled plugin binary in the field `plugin`.
```
maha-hajja, line 55:
```suggestion
can create a connector using the connectors API, and specify the path to the compiled plugin binary in the field `plugin`.
```

README.md
```diff
@@ -1,3 +1,77 @@
...
+
+Here is an example request to `/POST /v1/connectors` (find more about the [Conduit API](https://github.com/conduitio/conduit#api)):
```
maha-hajja, line 57:
```suggestion
Here is an example request to `POST /v1/connectors` (find more about the [Conduit API](https://github.com/conduitio/conduit#api)):
```

destination.go
```diff
@@ -71,8 +71,9 @@ type Destination interface {
 	Write(context.Context, Record) error
 
 	// Flush signals the plugin it should flush any cached records and call all
-	// outstanding AckFunc functions. No more calls to Write will be issued
-	// after Flush is called.
+	// outstanding AckFunc functions. This function needs to be implemented only
+	// if the struct also implements WriteAsync. No more calls to WriteAsync
```
maha-hajja, line 75:
```suggestion
	// if the struct implements WriteAsync. No more calls to WriteAsync
```

README.md
lovromazgon: Done, wdyt?

ahmeroxa:
LGTM 👍

README.md
maha-hajja: looks good 👍

README.md
```diff
@@ -1,3 +1,77 @@
 # Conduit Connector Plugin SDK
```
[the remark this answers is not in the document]
lovromazgon: I changed it to use only "plugin".

record.go
```diff
@@ -40,23 +40,36 @@ type Record struct {
 	Payload Data
 }
 
+// Position is a unique identifier for a record. It is the responsibility of the
+// Source to choose and assign record positions, it can freely choose a format
+// that makes sense and contains everything needed to restart a pipeline at a
+// certain position.
 type Position []byte
```
hariso, line 47: Maybe it would make sense, to bring this sentence into the first plan: "contains everything needed to restart a pipeline at a certain position." Positions being unique IDs of records does not necessarily mean that they can be used to restart a pipeline from a position.

hariso (approved):
LGTM! I left a tiny comment about a comment, certainly not a blocker.

record.go
lovromazgon: I spent some time trying to restructure the comment but I can't come up with anything without overcomplicating it. Besides, it is 4 lines, I think a developer should be capable of reading it in one go 😛  There is another comment explaining this more specifically on [`Source.Open`](https://github.com/ConduitIO/conduit-plugin-sdk/blob/db79c240b1faaacc94e9687d5468693deb0e8c07/source.go#L38-L42), that's where the logic is important after all.

That said, I welcome proposals if you already have a better comment in mind!

hariso: > reading it in one go
Pun intended? : )

But yeah, I definitely agree the comment is good, concise, I was thinking about defining a position primarily as something which is used in a source, to start reading from it. IIUC, that is the most important thing about it. Positions being unique IDs may not be enough.

lovromazgon:
@neovintage @anaptfox any more comments? Can we merge this?

neovintage:
@lovromazgon :shipit:

PR #2, opened 2022-03-01
Close context in Source.Open when plugin stops

lovromazgon:
Moving commit from Conduit to the SDK: https://github.com/ConduitIO/conduit/pull/230/commits/f8d99d9367473244325b624dfbf7ed9a35427787

This change makes sure that the context passed to `Open` will be closed once `Stop` is called. This is useful because `Open` might be used to spawn background jobs that need to be notified when the plugin should stop. Otherwise the user would need to detect it in the `Read` function.

### DIFF
--- a/source.go
+++ b/source.go
@@ -37,7 +37,8 @@ type Source interface {
 
 	// Open is called after Configure to signal the plugin it can prepare to
 	// start producing records. If needed, the plugin should open connections in
-	// this function.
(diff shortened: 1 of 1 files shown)


hariso (approved):
LGTM! It would be great if it would be possible to have a unit test for this.

lovromazgon:
Yeah, for this and for everything else in this repo 🙈

PR #4, opened 2022-03-21
Ensure context in Open stays open

lovromazgon:
The context passed to the plugin in `Open` should stay open even after the function returns because the context can be used to spawn background jobs that fetch records.


source.go
```diff
@@ -107,8 +109,25 @@ func (a *sourcePluginAdapter) Configure(ctx context.Context, req cpluginv1.Sourc
 }
 
 func (a *sourcePluginAdapter) Start(ctx context.Context, req cpluginv1.SourceStartRequest) (cpluginv1.SourceStartResponse, error) {
-	ctx, a.openCancel = context.WithCancel(ctx)
-	err := a.impl.Open(ctx, req.Position)
+	// create a new context, so we can control when it's canceled
+	ctxOpen := context.Background()
+	ctxOpen, a.openCancel = context.WithCancel(ctxOpen)
```
hariso, line 114: I'm wondering, why don't we simply do `ctxOpen, a.openCancel = context.WithCancel(context.Background())`?

Or maybe even `ctxOpen, a.openCancel = context.WithCancel(ctx)`?

lovromazgon: First option won't work unless you declare `ctxOpen` a line before with `var ctxOpen context.Context`, not really an improvement IMO.

Second option is essentially what we did before but it had a flaw - the context could be cancelled even after `Open` returns (and it actually happens), so that context is then not suitable for spawning long-running goroutines.

hariso: Thanks! TIL a bit more about multiple short var declarations.: )

Issue #5, opened 2022-03-31
Bug: Logger does not work in Open


lovromazgon:
In PR https://github.com/ConduitIO/conduit-connector-sdk/pull/4 the context passed to `Open` was replaced with a new context so that we can control how long it stays open. The problem is that this approach doesn't forward the context values that might be present in the original context, like for example the logger.

We should probably explore another approach like detaching the parent cancellation from the context instead of creating a fresh context (see [this thread](https://github.com/golang/go/issues/40221) for more info).

PR #6, opened 2022-03-31
Detach context

lovromazgon:
This PR creates a `DetachContext` function which is taken from https://cs.opensource.google/go/x/tools/+/refs/tags/v0.1.10:internal/xcontext/xcontext.go and improved slightly. A test is added to ensure the logger is attached to the context in `Open`.

Fixes https://github.com/ConduitIO/conduit-connector-sdk/issues/5.

### DIFF
--- a/internal/context.go
+++ b/internal/context.go
@@ -1,0 +1,32 @@
+// Copyright © 2022 Meroxa, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


internal/context.go
```diff
@@ -0,0 +1,32 @@
...
+// limitations under the License.
+
+package internal
+
+import (
+	"context"
+	"time"
+)
+
+// Borrowed from https://cs.opensource.google/go/x/tools/+/refs/tags/v0.1.10:internal/xcontext/xcontext.go
```
hariso, line 22: Do we then need to include the copyright notice from xcontext.go?

hariso (approved):
LGTM! I had one comment about a copyright notice, which may not be an issue,

internal/context.go
lovromazgon: I'm honestly not sure. See, it's not copied 1:1, it's mostly the idea that is taken from there but adjusted slightly (parent context is not a named field, the function name is different). I added the link to the original to at least give the original authors _some_ credit. But honestly, if I were to tackle this problem myself I would have written it the same way (it's just a few lines, hard to do things in a wildly different way).

That said, better safe than sorry, I will add the disclaimer to the top of the file.

PR #7, opened 2022-03-31
Wait for acks in destination stop

lovromazgon:
Changes logic to be more intuitive, now the SDK makes sure that `Stop` will wait for `Run` to return before it exits. It will wait for at most 1 minute. This logic was previously in `Teardown` but it makes more sense to wait for run to stop before any remaining records are flushed, otherwise more records might still come in.

This was discovered while working on https://github.com/ConduitIO/conduit/issues/259.

Tests are missing, although it was tested manually with Conduit integration tests and the file connector.

### DIFF
--- a/destination.go
+++ b/destination.go
@@ -109,6 +109,8 @@ type destinationPluginAdapter struct {
 	impl       Destination
 	wgAckFuncs sync.WaitGroup
 	isAsync    bool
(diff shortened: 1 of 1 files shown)


hariso (approved):
:muscle:

hariso (approved):

PR #14, opened 2022-04-13
Acceptance tests

lovromazgon:
### Description

Fixes https://github.com/ConduitIO/conduit/issues/38

This adds acceptance tests to the SDK which can be called from a connector to make sure it fulfills expectations.

Besides acceptance tests this also introduces the struct `Connector` which ties together the functions we had to manually gather before (constructors for creating specifications, source and destination). It's a minor improvement that makes the code nicer and safer since the connector can be defined globally once and used in the entry point as the argument to `Serve` as well as the argument to the acceptance test driver config.

For example, we can run the acceptance tests for the Kafka connector by creating `acceptance_test.go` with the following test (the code needs to be slightly adjusted to pass, the tests already found a couple of issues):

```
package kafka

import (
	"testing"

	sdk "github.com/conduitio/conduit-connector-sdk"
	"github.com/google/uuid"
)

func TestAcceptance(t *testing.T) {
	cfg := testConfigMap()

	sdk.AcceptanceTest(t, sdk.ConfigurableAcceptanceTestDriver{
		Config: sdk.ConfigurableAcceptanceTestDriverConfig{
			Connector: sdk.Connector{ // Note that this variable should rather be created globally in `connector.go`
				NewSpecification: Specification,
				NewSource:        NewSource,
				NewDestination:   NewDestination,
			},
			SourceConfig:      cfg,
			DestinationConfig: cfg,

			BeforeTest: func(t *testing.T) {
				// use new topic for each test
				cfg[Topic] = "test-topic-" + uuid.NewString()
			},
		},
	})
}
```

Some linters are disabled on file `acceptance_testing.go` since it's essentially a test file.

There are some tests still missing, I'll keep working on them but they shouldn't affect existing tests so that shouldn't block the review. Here's a list of tests I identified (please let me know if I forgot any important test so we add it to the list!):

- [x] source
  - [X] configure
    - [X] required parameters
    - [X] teardown after configure
  - [x] open
    - [x] resume at position
  - [X] read
    - [X] successful snapshot
    - [X] successful CDC read
    - [X] timeout / cancelled context
    - [X] goroutine leak
    - [X] unique position
  - [X] ack (part of resuming at position)
  - [X] stop (closed context in read tests)
  - [X] teardown (implicitly in configure/read tests)
- [ ] destination
  - [X] configure
    - [X] required parameters
    - [X] teardown after configure
  - [X] write
    - [X] successful write
    - [X] goroutine leak
  - [ ] async write
    - [X] successful write
    - [X] goroutine leak
    - [X] ack after flush
    - [ ] ack after X seconds (after I thought about it, we are not going to test this, instead we should add this functionality to the SDK - `Flush` should be called every X seconds if records were received, or after every Yth record)
    - [ ] ack after X records (not going to test this, see previous point for reasoning)
    - [ ] ack error
  - [X] stop (closed context in write tests)
  - [X] teardown (implicitly in configure/write tests)


dylanlott:
Looking good so far! 🔌 ✅ Re: Missing tests cases - Should we add a test case for idempotent `Teardown`'s? is that a standard we want to enforce at some level?

lovromazgon:
Re idempotent `Teardown` it's obviously nice, but not a requirement IMO, we are only ever going to call it once. What we _do_ test on the other hand is that `Teardown` can be called after `Configure` without a call to any other function in between.

dylanlott (approved):
LGTM. I was able to get an acceptance test running against the Postgres connector. Overall, I think this is a good approach that allows us fine-grained external testing of connector behaviors. The interface for the test driver should allow us a ton of flexibility to create different test harnesses. The goleak test output can be difficult to interpret. It might be worth looking into what tooling we could build to make goleak test output easier to read. [Further reading.](https://github.com/uber-go/goleak#determine-source-of-package-leaks)

serve.go
```diff
@@ -15,6 +15,7 @@
 package sdk
 
 import (
+	"errors"
```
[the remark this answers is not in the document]
lovromazgon: `cerrors` is a package in Conduit, the SDK does not depend on Conduit.

acceptance_testing.go
```diff
@@ -0,0 +1,834 @@
...
+		defer cancel()
+
+		got, err := source.Read(readCtx)
+		is.NoErr(err)
+
+		err = source.Ack(ctx, got.Position)
+		is.NoErr(err)
+
+		want[i].Position = got.Position   // position can't be determined in advance
+		want[i].CreatedAt = got.CreatedAt // created at can't be determined in advance
```
hariso, line 495: So this implicitly also asserts that records are read in the same order in which they were originally written to the source system?

acceptance_testing.go
```diff
@@ -0,0 +1,834 @@
...
+			// TODO ensure position is unique
+			// TODO ensure we check metadata and mark it as successful or skipped if it's not (don't fail test because of metadata)
+			is.Equal(want[i], got)
+		}
+	})
+
+	// while connector is running write more data and make sure the connector
+	// detects it
+	want = a.generateRecords(20)
+	a.driver.Write(t, &want)
```
hariso, line 548: IIUC, this basically limits the detected changes to inserts only?

acceptance_testing.go
```diff
@@ -0,0 +1,834 @@
...
+		is.Equal(len(got), len(want)) // destination didn't write expected number of records
+		for i := range want {
+			want[i].Position = got[i].Position   // position can't be determined in advance
+			want[i].CreatedAt = got[i].CreatedAt // created at can't be determined in advance
+
+			is.Equal(want[i], got[i])
+		}
+	})
+
+	is.True((len(gotSynchronous) > 0) != (len(gotAsynchronous) > 0)) // either Write or WriteAsync should be implemented and working (not both)
```
hariso, line 715: Can we somehow enforce this in the SDK? It would also be good that we mention this in the docs.

acceptance_testing.go
```diff
@@ -0,0 +1,834 @@
...
+		Metadata:  nil,                        // TODO metadata is optional so don't enforce it to be written, still create it
+		CreatedAt: time.Unix(0, a.rand.Int63()),
+		Key:       RawData(a.randString(32)), // TODO try structured data as well
+		Payload:   RawData(a.randString(32)), // TODO try structured data as well
+	}
+}
+
+// randString generates a random string of length n.
+// (source: https://stackoverflow.com/a/31832326)
+func (a acceptanceTest) randString(n int) string {
```
hariso, line 812: It looks like we're always calling randString with the same length parameter, so maybe we can simply use a UUID here.

acceptance_testing.go
```diff
@@ -0,0 +1,834 @@
...
+	// suppress false positive goroutine leaks.
+	GoleakOptions(*testing.T) []goleak.Option
+
+	// Write receives a slice of records that should be prepared in the 3rd
+	// party system so that the source will read them. The slice will be used to
+	// verify the source connector can successfully execute reads.
+	// It is discouraged for the driver to change the record slice, unless there
+	// is no way to write the records to the 3rd party system, then the slice
+	// should be modified to reflect the expected records a source should read.
+	Write(*testing.T, *[]Record)
```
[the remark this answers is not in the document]
lovromazgon: Thanks for the suggestion, I've been struggling with choosing a good name for this, maybe I slimmed it down too much in the end. I'll change it to `WriteToSource` and `ReadFromDestination`.

acceptance_testing.go
lovromazgon: That's right, right now we test inserts only. We might want to check updates and deletes as well, the problem is we don't have a way to represent updated or deleted records right now, but this should be doable once we switch to the OpenCDC record.

acceptance_testing.go
lovromazgon: If by "enforce" you mean that it would produce a compile-time error, I don't see how we could do that. I'll add a comment to the docs about this 👍

acceptance_testing.go
lovromazgon: That could work, although the generated string would be somewhat limited (only `0-9`, `a-f` and `-`). I wanted to generate a string that's a bit more versatile to make sure the connector can handle it. Maybe we could go a step further and for raw data create just random bytes that are not necessarily a valid UTF-8 string (raw data can theoretically be _anything_ after all).

This reminded me of another problem related to generating data - some connectors might not support raw data at all (e.g. postgres right now works only with structured data or with JSON raw data). That might be completely valid behavior, in that case the user would have to use transforms to create structured data (e.g. use the [hoist transform](https://github.com/ConduitIO/conduit/blob/85f46e4db4041403c4bc8af9c2487c5256bc66f0/pkg/processor/transform/txfbuiltin/hoistfield.go#L49-L50)). So maybe we could give the driver the control over how records are generated and by default generate both, raw and structured data. If the connector can't handle it, the user can restrict it to either one or the other, or overwrite the method entirely and generate their own records. WDYT?

acceptance_testing.go
```diff
@@ -0,0 +1,836 @@
...
+	AfterTest func(t *testing.T)
+
+	// GoleakOptions will be applied to goleak.VerifyNone. Can be used to
+	// suppress false positive goroutine leaks.
+	GoleakOptions []goleak.Option
+
+	// Skip lets the caller define if any tests should be skipped (useful for
+	// skipping destination/source tests if the connector only implements one
+	// side of the connector)
+	Skip []string
```
maha-hajja, line 133: shouldn't we mention here that the string is a regex?

acceptance_testing.go
lovromazgon: I added a warning on top of `AcceptanceTest` since it seems pretty important to sort this out before we can actually rely on this test.

acceptance_testing.go
lovromazgon: Will update the doc! Actually we need to remove the additional tip about skipping destination/source tests, because that's not needed anymore, the test itself will determine to skip source/destination tests if the connector does not implement it.

acceptance_testing.go
hariso: I'm wondering if it's too strong of a requirement or if certain connectors will ever meet it. E.g. the S3 connector: from what I recall, we have no way of knowing the insertion order. When listing S3 objects, they are sorted by key.

acceptance_testing.go
hariso: Cool, that makes sense to me. :+1:

acceptance_testing.go
hariso: Or runtime (e.g. when writing the first record, but that is likely not a good experience). 

Not related to this, it may make sense to split test into three: one for sync, one for async, and one checking if exactly one of the methods is implemented.

acceptance_testing.go
hariso: I agree, it would be good if the driver would generate the records, due to not only types of payloads, but maybe even contents.

> Although this opens another problem - if we want to test that a destination connector knows how to write structured data, but the source only produces raw data, how do we assert we got back the expected records? sweat_smile This needs to be thought through in a separate issue.

Maybe we won't ever face such a situation, since I believe that usually a source will produce the type of payload which makes most sense for a system (S3 - raw, Pg - struct), and then the destination will require again the type of payload which makes sense for that system. 

Or maybe we need to decouple sources and destinations in the tests, so that we don't use sources to check if a destination write as OK (for example).

But yeah, I agree that this needs more thought in a separate issue. :+1:

.github/ISSUE_TEMPLATE/2-bug.yml
```diff
@@ -1,4 +1,4 @@
-name: Bug
+name: 🐛 Bug
```
hariso, line 1: I was 99% sure this bug was green.:D

acceptance_testing.go
lovromazgon: Good point 🤔 IMO it would feel wrong to just accept any order though, it feels too vague if we let the source produce just a collection of records in any order. I'm using the word "feels", I have no hard data to back that up 🙃 Other than maybe records could reference each other, so order is important in that case (e.g. record B references record A, so ideally A is produced before B).

I thought about if there's a middle ground. We could say the connector should be deterministic, in the sense that it always produces the same order. It could be _any_ order, as long as it's always the same order. This is already achievable by writing your own driver and changing the order of the records in the expectation:
```
type AcceptanceTestDriver struct {
	sdk.ConfigurableAcceptanceTestDriver
}

func (d AcceptanceTestDriver) WriteToSource(t *testing.T, records *[]sdk.Record) {
	// sort records by key ...
	d.ConfigurableAcceptanceTestDriver.WriteToSource(t, records)
}
```

Thoughts?

acceptance_testing.go
lovromazgon: Split into `TestDestination_WriteOrWriteAsync`, `TestDestination_Write_Success` and `TestDestination_WriteAsync_Success` 👍

acceptance_testing.go
lovromazgon: I just noticed that it's not possible to do the same in `ReadFromDestination`, we would have to pass in the slice of the expectations so it could be changed.

lovromazgon: Alright, so I changed the signature of `ReadFromDestination` and `WriteToSource` so the parameter is a slice of records that should be read/written and the returned slice is what actually was read/written. This means that the order of the records can be changed as the driver sees fit.

hariso: Sounds good! To the order: it _feels_ (there it is again), that the order is important in a minority of cases, and it's probably limited to those cases where there is some kind of relationship between records. When it comes to at least some sources I've worked with (RDBMs, Kafka) you can't always guarantee the ordering, so maybe a default of no ordering assumed in tests works best.

lovromazgon: The order of records shouldn't play a role anymore now!

hariso: > The order of records shouldn't play a role anymore now!

Anarchy loading... :D

hariso (approved):
:rocket:

PR #15, opened 2022-04-26
adds deadline to source teardown

dylanlott:
### Description

Adds a deadline to a source's Teardown to put a time limit on how long the connector will wait for Run to stop and return.

Partially addresses Conduit issue #185.

### Quick checks:

  update/change.


lovromazgon (commented):
This doesn't really address the issue. The `Teardown` function is not the problem, but rather `a.t.Wait()` which will block forever if the stream doesn't get closed.

That said, I think we should postpone this issue because while writing the pipeline semantics document I spotted a bigger issue - the SDK right now expects to receive all acknowledgments before it stops the stream, but this might not be the case if a message gets nacked. We will have to rethink the `Stop`/`Teardown` functions a bit to make this work and make Conduit in charge of determining when to close the stream. I'll open an issue after I'm done with the doc describing the expected behavior.

PR #20, opened 2022-05-20
 Acceptance Tests: Move generating of records to driver

### LINKED ISSUE #16: Acceptance Tests: Move generating of records to driver
We should move the method `acceptanceTest.generateRecord` to `AcceptanceTestDriver`. This will allow us to configure how we want the record to be generated or let the user overwrite the method to generate their own records if needed.

See this comment for more info: https://github.com/ConduitIO/conduit-connector-sdk/pull/14#discussion_r866781828

hariso:
### Description

Move test record generation to the acceptance test driver. Depends on:
* #14 

This is done by moving the `generateRecord` method to the driver. Another option was to move `generateRecords`, but it felt a slightly better experience not to write a loop yourself.: )

Fixes #16 

### Quick checks:

  update/change.


lovromazgon (approved):
Great ✨

PR #22, opened 2022-05-31
Destination stop semantics

lovromazgon:
### Description

This changes the semantics of a graceful destination stop. With the information about the last record position we can now determine when we received all records in the plugin and safely execute `Flush`, knowing that it will flush all records (no race condition). This PR also adds a bunch of tests for the destination.

Fixes https://github.com/ConduitIO/conduit/issues/391.

~Depends on https://github.com/ConduitIO/conduit-connector-protocol/pull/6.~

### Quick checks:

  update/change.


internal/holder.go
```diff
@@ -0,0 +1,120 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"sync"
+)
+
+// Holder holds a reference to a value. Multiple goroutines are able to put or
+// retrieve the value into/from the Holder, as well as wait for a certain value
+// to be put into the Holder.
+type Holder[T any] struct {
```
lovromazgon, line 27: If you come up with a better name for this struct, please let me know!

dylanlott (approved):
Nice use of generics for saving the last position. Looks good to me.

hariso (approved):

PR #30, opened 2022-07-15
Acceptance tests doc

lovromazgon:
### Description

This PR creates a document that gathers my research into acceptance tests. This was triggered by the integration of OpenCDC records into the SDK. It is a dump of my thoughts on acceptance tests that should serve as a basis for discussion around this topic. If/when we agree on the approach, we can take information from this document to create specific issues around implementing acceptance tests.

PR #31, opened 2022-07-18
OpenCDC integration

lovromazgon:
### Description

This PR integrates the new OpenCDC record into the SDK. The acceptance tests are adjusted minimally just to make them work, but we will need to revisit them and adjust expectations for different kinds of connectors (see #30 for more info).

Depends on https://github.com/ConduitIO/conduit-connector-protocol/pull/9 ~(needs to be merged and we need to update the connector protocol version)~.

Fixes https://github.com/ConduitIO/conduit/issues/510.
Fixes https://github.com/ConduitIO/conduit/issues/223 (added utility functions for creating a record, they will add the metadata field `opencdc.readAt`.

### Quick checks:

  update/change.


acceptance_testing.go
```diff
@@ -1074,27 +1086,65 @@ func (a acceptanceTest) isEqualRecords(is *is.I, want, got []Record) {
...
-	is.Equal(len(got), len(gotMap)) // record payloads are not unique
+	for i := range want {
+		a.isEqualRecord(is, want[i], got[i])
+	}
+}
 
-	for key, wantRec := range wantMap {
-		gotRec, ok := gotMap[key]
-		is.True(ok) // expected record not found
+func (a acceptanceTest) sortMatchingRecords(want, got []Record) {
```
hariso, line 1115: Nitpick: would be nice to briefly explain what sortMatchingRecords is about, especially since it's "flexible" in sorting.

metadata.go
```diff
@@ -0,0 +1,147 @@
...
+// SetOpenCDCVersion sets the metadata value for key MetadataVersion to the
+// current version of OpenCDC used. This
+func (MetadataUtil) SetOpenCDCVersion(metadata map[string]string) {
+	metadata[MetadataOpenCDCVersion] = cpluginv1.OpenCDCVersion
+}
+
+// GetOpenCDCVersion returns the value for key
+// MetadataOpenCDCVersion. If the value is does not exist or is empty the
+// function returns ErrMetadataFieldNotFound.
+func (u MetadataUtil) GetOpenCDCVersion(metadata map[string]string) (string, error) {
```
hariso, line 61: Java <3 :D

metadata.go
```diff
@@ -0,0 +1,147 @@
...
+func (MetadataUtil) SetCreatedAt(metadata map[string]string, createdAt time.Time) {
+	metadata[MetadataCreatedAt] = strconv.FormatInt(createdAt.UnixNano(), 10)
+}
+
+// GetReadAt parses the value for key MetadataReadAt as a unix
+// timestamp. If the value does not exist or the value is empty the function
+// returns ErrMetadataFieldNotFound. If the value is not a valid unix timestamp
+// in nanoseconds the function returns an error.
+func (MetadataUtil) GetReadAt(metadata map[string]string) (time.Time, error) {
+	raw := metadata[MetadataReadAt]
```
hariso, line 94: It looks like the code we have in GetReadAt can be reused in GetCreatedAt.

record_test.go
```diff
@@ -0,0 +1,46 @@
...
+				"foo": "bar",
+				"baz": "qux",
+			},
+		},
+	}
+
+	want := `{"position":"Zm9v","operation":"create","metadata":{"conduit.plugin.name":"example","opencdc.version":"v1"},"key":"YmFy","payload":{"before":null,"after":{"baz":"qux","foo":"bar"}}}`
+
+	got := string(r.Bytes())
+	is.Equal(got, want)
```
hariso, line 45: Nitpick: we may want to verify the OpenCDC version is not left behind in the metadata.

record_test.go
```diff
@@ -0,0 +1,46 @@
...
+		Payload: Change{
+			Before: nil,
+			After: StructuredData{
+				"foo": "bar",
+				"baz": "qux",
+			},
+		},
+	}
+
+	want := `{"position":"Zm9v","operation":"create","metadata":{"conduit.plugin.name":"example","opencdc.version":"v1"},"key":"YmFy","payload":{"before":null,"after":{"baz":"qux","foo":"bar"}}}`
```
hariso, line 42: Nitpick: It might be nice to add some new lines in this string, so that it's easier to to skim over it.

metadata.go
```diff
@@ -0,0 +1,147 @@
...
+	// MetadataConduitPluginName is a Record.Metadata key for the name of the
+	// plugin that created this record.
+	MetadataConduitPluginName = cpluginv1.MetadataConduitPluginName
+	// MetadataConduitPluginVersion is a Record.Metadata key for the version of
+	// the plugin that created this record.
+	MetadataConduitPluginVersion = cpluginv1.MetadataConduitPluginVersion
+)
+
+// MetadataUtil provides utility methods for reading and writing metadata.
+type MetadataUtil struct{}
```
hariso, line 50: Would it make sense to have a metadata type which would have map[string]string as the undelying type, and which we would use in Record? That way, we could have the method below directly on metadata, and it would be easier to call them. E.g.
```
type metadata map[string]string

func (m metadata) SetFoo() {
	m["foo"] = "bar"
}

type Record struct {
    m metadata
    ...
}
``` 
I'm not quite sure how would it affect serialization though.

hariso (approved):
Looks good to me! I left a few nitpicks, feel free to merge without those.: )

metadata.go
lovromazgon: Good idea, I tried it out and I think it looks nice 👍 lmk what you think about about the last commit

hariso: Looks good!

record_test.go
lovromazgon: I'll skip this one, we would need to do some magic to compare the expectations then, either parse them before comparing or reformat them. Not worth it at the moment IMO.

record_test.go
lovromazgon: Done!

record_test.go
hariso: Ah, right, the new lines would be serialized as well.

metadata.go
lovromazgon: I'll skip this as well, let's think about it when/if we have another metadata field that contains unix nanos. I think DRY is an overrated principle that is abused in cases like this, we can discuss it some time 🙂

acceptance_testing.go
lovromazgon: Done!

hariso (approved):

metadata.go
hariso: I do agree that it's an overrated principle. The reason why I think it would be good to apply in cases like these is because the logic for both is the same, and will likely stay the same for these methods (as it's the same type). But not a big thing anyway. 

> we can discuss it some time slightly_smiling_face

I'll be glad!

PR #40, opened 2022-11-15
ability to add validations to parameters

maha-hajja:
### Description

This PR adds a `Validation` interface to the SDK, with all the validation types.
this will give the ability for the developer to add validations as so:

```
func (s *Source) Parameters() map[string]sdk.Parameter {
	return map[string]sdk.Parameter{
		"param: {
			Default:     "",
			Required:    true,
                         Type: sdk.ParameterTypeNumber,
			Validations: []sdk.Validation{
					sdk.ValidationRequired{},
					sdk.ValidationGreaterThan{Value: 1},
					sdk.ValidationLessThan{Value: 4.1},
					sdk.ValidationInclusion{List: []string{"2", "3"}},
					sdk.ValidationExclusion{List: []string{"5", "6"}},
					sdk.ValidationRegex{Pattern: "*"},
				},
		},
}
```

the implementation for the validations is still in progress, these are marked as "TBD" and will be addressed in the next PR for https://github.com/ConduitIO/conduit/issues/695 

Fixes #https://github.com/ConduitIO/conduit/issues/694

### Quick checks:

  update/change.


validation.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/conduitio/conduit-connector-protocol/cpluginv1"
+)
+
+const (
+	ValidationTypeRequired ValidationType = iota + 1
+	ValidationTypeGreaterThan
+	ValidationTypeLessThan
+	ValidationTypeInclusion
+	ValidationTypeExclusion
+	ValidationTypeRegex
```
lovromazgon, line 30: IIUC the connector developer won't use these validation types, right? If it's only used internally we should make them private, or maybe even remove it completely and use the `cpluginv1` constants directly (no need for the abstraction if it's only internal).

validation.go
```diff
@@ -0,0 +1,176 @@
...
+	_ = cTypes[int(ParameterTypeNumber)-int(cpluginv1.ParameterTypeNumber)]
+	_ = cTypes[int(ParameterTypeBool)-int(cpluginv1.ParameterTypeBool)]
+	_ = cTypes[int(ParameterTypeFile)-int(cpluginv1.ParameterTypeFile)]
+	_ = cTypes[int(ParameterTypeDuration)-int(cpluginv1.ParameterTypeDuration)]
+}
+
+type Validation interface {
+	validate(key string, value string) error
+	vType() ValidationType
+	value() string
```
lovromazgon, line 63: We need this to create a `cpluginv1.ParameterValidation`, think about grouping them together into one function, for example:
```suggestion
	toCPluginv1() cpluginv1.ParameterValidation
```

validation.go
```diff
@@ -0,0 +1,176 @@
...
+
+	_ = cTypes[int(ParameterTypeString)-int(cpluginv1.ParameterTypeString)]
+	_ = cTypes[int(ParameterTypeNumber)-int(cpluginv1.ParameterTypeNumber)]
+	_ = cTypes[int(ParameterTypeBool)-int(cpluginv1.ParameterTypeBool)]
+	_ = cTypes[int(ParameterTypeFile)-int(cpluginv1.ParameterTypeFile)]
+	_ = cTypes[int(ParameterTypeDuration)-int(cpluginv1.ParameterTypeDuration)]
+}
+
+type Validation interface {
+	validate(key string, value string) error
```
lovromazgon, line 61: Why do we need the `key`? If it's for creating the error message, I'd rather we wrap the error in the caller and add the key to the error there, this will ensure all error messages have the same format.

validation.go
```diff
@@ -0,0 +1,176 @@
...
+
+func (v ValidationRegex) vType() ValidationType {
+	return ValidationTypeRegex
+}
+
+func (v ValidationRegex) value() string {
+	return v.Pattern
+}
+
+func ConvertValidations(validations []Validation) []cpluginv1.ParameterValidation {
```
lovromazgon, line 167: This should be private.

validation.go
maha-hajja: I actually use vType() when validating the parameter, if the parameter has a validation lessThan(2) but it's not required, then I would skip the other validations. But this can still be done by converting to cpluginv1 and checking the type.

specifier.go
```diff
@@ -43,11 +43,14 @@ type Parameter struct {
 	// Default is the default value of the parameter, if any.
 	Default string
 	// Required controls if the parameter will be shown as required or optional.
-	// Note that no validation will be added to the field because of this, this
-	// is purely for informational purposes.
+	// Deprecated: add a validation of type TYPE_REQUIRED to the validations list.
```
lovromazgon, line 46:
```suggestion
	// Deprecated: add ValidationRequired to Parameter.Validations instead.
```

lovromazgon (approved):
LGTM!

validation.go
lovromazgon: I see, so `ValidationRequired` has a bit of a special status. I think we can get away without knowing the validation type though, we can check the Go type instead.
```
for _,v := range sliceOfValidations {
	_, isRequired := v.(ValidationRequired)
	if isRequired {
		// this parameter is required
	}
}
```

PR #41, opened 2022-11-22
apply parameter validations in the configure method

maha-hajja:
### Description

added a function that:
* applies default values before doing validations
* finds any unrecognized parameters (could be a typo by the user)
* validates the type of the parameters (number, bool, file, ..)
* validates the value of the parameter depending on the validations provided (less-than, regex, ..)
* returns multierror of all the errors occurred

added tests for all the cases.

this function is called in the `Configure` method for the source and destination.

Fixes #https://github.com/ConduitIO/conduit/issues/695

### Quick checks:

  update/change.


validation.go
```diff
@@ -51,7 +67,9 @@ type ValidationRequired struct {
 }
 
 func (v ValidationRequired) validate(value string) error {
-	// TBD
+	if value == "" {
```
hariso, line 70: What if the string is all white spaces?

validation.go
```diff
@@ -149,3 +190,71 @@ func convertValidations(validations []Validation) []cpluginv1.ParameterValidatio
...
+		if err != nil {
+			multiErr = multierr.Append(multiErr, fmt.Errorf("error validating %q: %w", key, err))
+		}
+	}
+	return multiErr
+}
+
+// assignParamDefaults fills any empty configuration with its assigned default value
+// returns an error if a parameter is not recognized.
+func assignParamDefaults(params map[string]Parameter, config map[string]string) error {
```
hariso, line 224: WDYT about not mutating the input configuration map and instead returning the supplemented map?

validation.go
```diff
@@ -149,3 +190,71 @@ func convertValidations(validations []Validation) []cpluginv1.ParameterValidatio
...
+		_, err := time.ParseDuration(value)
+		if err != nil {
+			return fmt.Errorf("%q: %q value is not a duration: %w", key, value, ErrInvalidParamType)
+		}
+	case ParameterTypeBool:
+		_, err := strconv.ParseBool(value)
+		if err != nil {
+			return fmt.Errorf("%q: %q value is not a boolean: %w", key, value, ErrInvalidParamType)
+		}
+		// type ParameterTypeFile and ParameterTypeString don't need type validations, since a file is an array of bytes,
```
hariso, line 256: If ParameterTypeFile represents a file, which is an array of bytes, and since all configuration values are string, how exactly should a file be encoded as a string?

validation.go
```diff
@@ -149,3 +190,71 @@ func convertValidations(validations []Validation) []cpluginv1.ParameterValidatio
...
+			config[key] = params[key].Default
+		}
+	}
+	return multiErr
+}
+
+// validateParamType validates that a parameter value is parsable to its assigned type.
+func validateParamType(p Parameter, key string, value string) error {
+	switch p.Type {
+	case ParameterTypeNumber:
```
hariso, line 240: WDYT about having different types for integers and floats?

validation.go
maha-hajja: either way is fine... is it a better convention or a preference?

validation.go
maha-hajja: I meant that an array of bytes is basically a string, so we can't validate that it's a file, it's just a string that doesn't make sense. so the UI would upload a file and convert it to string.
The whole file parameter is still not implemented in the UI and this would be used in the future.

validation.go
maha-hajja: hmm it makes sense, a number works fine both ways, but I can see how having two types is better for the developer to add constraints on the user's configurations.
that would need a change for the connector protocol again, the number type was added recently and basically not used by anyone.. IIUC this mean we can replace it with another type without causing any backward compatibility problems. right? 
@lovromazgon WDYT?

validation.go
maha-hajja: I thought that the grcp request will take care of there for some reason.. but just tested it out and it doesn't :D 
I'll fix it.. thanks!

validation.go
hariso: There are fields that can only be integers (e.g. timestamps, batch sizes, etc.), and if you by mistake enter a float, the validation will pass, even though it's not really correct.

lovromazgon: > this mean we can replace it with another type without causing any backward compatibility problems

Yes, as long as the "string" type stays untouched this shouldn't create any issues. Let's do it now, after the next release it won't be so easy anymore.

validation.go
hariso: What I meant is that the string is an encoding of the input file. It may be a Base64-encoded version of the file for example. But since we don't support it now, I don't think it makes sense to "assign" meaning to it yet.

validation.go
hariso: IMHO, mutating parameters can lead to surprise for the caller of a method, but I also acknowledge that it's convenient. At least I'm leaning towards not mutating input parameters, but I wouldn't say that's a convention that should be followed.

maha-hajja: done

metadata.go
```diff
@@ -38,12 +38,20 @@ const (
...
+	// version of the source plugin that created this record.
+	MetadataConduitSourcePluginVersion = cpluginv1.MetadataConduitSourcePluginVersion
+	// MetadataConduitDestinationPluginName is a Record.Metadata key for the
+	// name of the destination plugin that has written this record
+	// (only available in records once they are written by a destination).
+	MetadataConduitDestinationPluginName = cpluginv1.MetadataConduitDestinationPluginName
+	// MetadataConduitDestinationPluginVersion is a Record.Metadata key for the
+	// version of the destination plugin that has written this record
+	// (only available in records once they are written by a destination).
+	MetadataConduitDestinationPluginVersion = cpluginv1.MetadataConduitDestinationPluginVersion
```
hariso, line 54: Do we have these documented somewhere?

validation.go
```diff
@@ -67,7 +87,13 @@ type ValidationLessThan struct {
...
 
 func (v ValidationLessThan) validate(value string) error {
-	// TBD
+	val, err := strconv.ParseFloat(value, 64)
+	if err != nil {
+		return fmt.Errorf("%q value should be a number: %w", value, ErrInvalidParamValue)
+	}
+	if !(val < v.Value) {
+		return fmt.Errorf("%q should be less than %.4f: %w", value, v.Value, ErrLessThanValidationFail)
+	}
```
hariso, line 96: Shouldn't we have different GT and LT validations for ints and floats, since the respective value ranges are different?

hariso (approved):
:rocket: LGTM! I left two minor comments, nothing blocking AFAIC.

metadata.go
maha-hajja: @lovromazgon is working on these changes

validation.go
hariso: @maha-hajja What I meant is that in this code we have in `ValidationLessThan` works on floats but since integers and floats have different value ranges, that won't always be correct. For example, what if the value cannot fit into an integer (e.g. too big)? The code above would pass (because the value can fit into a float). But, as I was typing this I realized that we have a type check which comes before this, so you can ignore this!

validation.go
```diff
@@ -149,3 +192,79 @@ func convertValidations(validations []Validation) []cpluginv1.ParameterValidatio
...
+// consists of a combination of errors from the configurations.
+func applyConfigValidations(params map[string]Parameter, config map[string]string) error {
+	// cfg is the config map with default values assigned
+	cfg, multiErr := assignParamDefaults(params, config)
+
+	for pKey, param := range params {
+		if len(strings.TrimSpace(cfg[pKey])) != 0 {
+			multiErr = multierr.Append(multiErr, validateParamType(param, pKey, cfg[pKey]))
+		}
+		multiErr = multierr.Append(multiErr, validateParamValue(pKey, cfg[pKey], param))
```
lovromazgon, line 206: My OCD kicks in when I see two functions with the same arguments but in a different order 😬 🙈

validation.go
```diff
@@ -149,3 +192,79 @@ func convertValidations(validations []Validation) []cpluginv1.ParameterValidatio
...
+	return output, multiErr
+}
+
+// validateParamType validates that a parameter value is parsable to its assigned type.
+func validateParamType(p Parameter, key string, value string) error {
+	switch p.Type {
+	case ParameterTypeInt:
+		_, err := strconv.Atoi(value)
+		if err != nil {
+			return fmt.Errorf("%q: %q value is not an integer: %w", key, value, ErrInvalidParamType)
```
lovromazgon, line 249: `validateParamValue` will create errors that are prefixed with `error validating %q: ...`, I think we should do the same here if the type is not correct.

validation.go
```diff
@@ -149,3 +192,79 @@ func convertValidations(validations []Validation) []cpluginv1.ParameterValidatio
 	}
 	return out
 }
+
+// ApplyConfigValidations a utility function that applies all the validations for parameters, return an error that
+// consists of a combination of errors from the configurations.
+func applyConfigValidations(params map[string]Parameter, config map[string]string) error {
```
lovromazgon, line 198: Last comment! 😄 I propose grouping these validation-related functions under a private struct, this way we wouldn't pollute the package with static functions. The type could even be a `map[string]Parameter`, then all functions will have access to the parameters already.
```
type validator map[string]Parameter

func (v validator) Validate(config map[string]string) error {
    ...
}

// we can use it like this
err := validator(source.Params()).Validate(config)
```

This is just one way to do it, but in any case I think creating a struct for this would make the code a bit nicer 😉

lovromazgon (commented):
Nice work! I like how we are able to figure out exactly what the main error was, this can come in handy if we want to return them in a more structured way or add error codes in the future 👍

My comments are mostly nitpicks, the important ones are about populating defaults and grouping validation functions in a struct to organize the code.

metadata.go
lovromazgon: They are documented here and in the [connector protocol](https://github.com/ConduitIO/conduit-connector-protocol/blob/main/cpluginv1/metadata.go) for now. But they are not used in any plugin yet afaik.

validation.go
```diff
@@ -131,7 +167,14 @@ type ValidationRegex struct {
 }
 
 func (v ValidationRegex) validate(value string) error {
-	// TBD
+	r, err := regexp.Compile(v.Pattern)
```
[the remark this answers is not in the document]
maha-hajja: you mean `*regexp.Regexp`instead of `string` right? that would mean the developer would have an extra step of creating the regex. but I agree this would be helpful with catching the invalid pattern early/

validation.go
```diff
@@ -193,27 +191,33 @@ func convertValidations(validations []Validation) []cpluginv1.ParameterValidatio
...
-		if len(strings.TrimSpace(cfg[pKey])) != 0 {
-			multiErr = multierr.Append(multiErr, validateParamType(param, pKey, cfg[pKey]))
+	var err error
+	for pKey := range v {
+		err = v.validateParamType(pKey, cfg[pKey])
+		multiErr = multierr.Append(multiErr, err)
+		// validate param value only if param type is valid
+		if err == nil {
+			multiErr = multierr.Append(multiErr, v.validateParamValue(pKey, cfg[pKey]))
 		}
```
lovromazgon, line 210: Just a nitpick - we should strive to keep the happy path to the left and indent error handling, if we follow this principle the code will be consistent and better readable. So here I would rather turn it around.
```suggestion
		err = v.validateParamType(pKey, cfg[pKey])
		if err != nil {
			// append error and continue with next parameter
			multiErr = multierr.Append(multiErr, err)
			continue
		}
		multiErr = multierr.Append(multiErr, v.validateParamValue(pKey, cfg[pKey]))
```

validation.go
```diff
@@ -222,46 +226,61 @@ func validateParamValue(key string, value string, param Parameter) error {
...
 			multiErr = multierr.Append(multiErr, fmt.Errorf("unrecognized parameter %q", key))
 			continue
 		}
+		// set config map and assign defaults
 		output[key] = val
-		if len(strings.TrimSpace(val)) == 0 {
-			output[key] = params[key].Default
+		if val == "" {
+			output[key] = v[key].Default
+		}
```
lovromazgon, line 247: I think we can simplify - if we remove this ...
```suggestion
```

validation.go
```diff
@@ -222,46 +226,61 @@ func validateParamValue(key string, value string, param Parameter) error {
...
+	}
+	// set defaults for unassigned params
+	for key, param := range v {
+		if _, ok := output[key]; !ok {
```
lovromazgon, line 251: ... and change this to check if the value is empty.
```suggestion
		if output[key] == "" {
```

validation.go
```diff
@@ -193,27 +191,33 @@ func convertValidations(validations []Validation) []cpluginv1.ParameterValidatio
...
 }
 
-// ApplyConfigValidations a utility function that applies all the validations for parameters, return an error that
+// private struct to group all validation functions
+type validator map[string]Parameter
+
+// applyConfigValidations is a utility function that applies all the validations for parameters, returns an error that
 // consists of a combination of errors from the configurations.
-func applyConfigValidations(params map[string]Parameter, config map[string]string) error {
+func (v validator) applyConfigValidations(config map[string]string) error {
```
lovromazgon, line 199: Now that this is a method on a private struct we can make it public - this makes it clear it is supposed to be called by some other code and not the validator, while other private methods are only internal helpers. And I think we can just call it `Validate` if you ask me, keep it simple 🙂 
```suggestion
func (v validator) Validate(config map[string]string) error {
```

validation.go
```diff
@@ -222,46 +226,61 @@ func validateParamValue(key string, value string, param Parameter) error {
...
 		}
 	}
 	return output, multiErr
 }
 
 // validateParamType validates that a parameter value is parsable to its assigned type.
-func validateParamType(p Parameter, key string, value string) error {
-	switch p.Type {
+func (v validator) validateParamType(key string, value string) error {
+	// empty value is valid for all types
```
lovromazgon, line 260: Good catch! 👍

lovromazgon (approved):
Very nice! 👍 🙌

validation.go
maha-hajja: writing the statement `if err==nil` honestly didn't feel right.. so yes I like this more!

validation.go
maha-hajja: wow that was something I wanted to do but forgot to for some reason 🥲 thanks for the suggestion

validation.go
maha-hajja: not really a catch.. it was checked in the function calling this one.. I just moved it 👀

PR #42, opened 2022-12-01
add ParseConfig function under Util

maha-hajja:
### Description

Related to #(https://github.com/ConduitIO/conduit/issues/696)

### Quick checks:

  update/change.

### DIFF
--- a/util.go
+++ b/util.go
@@ -14,14 +14,25 @@
 
 package sdk
 
-import "fmt"
+import (
+	"fmt"
+
+	"github.com/mitchellh/mapstructure"
+)
 
 // Util provides utilities for implementing connectors.
-var Util struct {
+var Util = struct {
 	// SourceUtil provides utility methods for implementing a source.
 	Source SourceUtil
 	// SourceUtil provides utility methods for implementing a destination.
 	Destination DestinationUtil
+	// ParseConfig provided to parse a config map into a struct
+	// Under the hood, this function uses the library mitchellh/mapstructure, so to rename a key, use
+	// the "mapstructure" tag and set a value directly. To embed structs, append ",squash" to your tag.
+	// for more details and docs, check https://pkg.go.dev/github.com/mitchellh/mapstructure
+	ParseConfig func(map[string]interface{}, interface{}) error
+}{
+	ParseConfig: parseConfig,
 }
 
 func mergeParameters(p1 map[string]Parameter, p2 map[string]Parameter) map[string]Parameter {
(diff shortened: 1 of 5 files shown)


util_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+	var got Person
+	err := Util.ParseConfig(input, &got)
+	is.NoErr(err)
+	is.Equal(want, got)
+}
+
+func TestParseConfig_All_Types(t *testing.T) {
+	is := is.New(t)
+	type Config struct {
```
lovromazgon, line 81: How about a slice? I'd expect to be able to parse a comma separated string into `[]string` (e.g. `a,b,c`). Bonus points if we can parse into `[]T` where `T` is any base type (that might be a bit harder though).

Another thing would be maps. For example, do we want to parse something like `key1:val1,key2:val2` into `map[string]string{"key1":"val1","key2":"val2"}`?

maha-hajja: @lovromazgon I don't think we ever had a config value as a map so far.. do you think we should support that too?

lovromazgon: We have a case in the generator where we do this ([link to a test](https://github.com/ConduitIO/conduit-connector-generator/blob/884dfe8d4dc4ed0c631830c7165d0c1934ab8254/config_test.go#L38)), but to be fair it's the only use case and I'm not sure there will be many more, so a map is not essential. A slice on the other hand would be useful IMO.

util.go
```diff
@@ -38,3 +49,21 @@ func mergeParameters(p1 map[string]Parameter, p2 map[string]Parameter) map[strin
 	}
 	return params
 }
+
+func parseConfig(raw map[string]string, config interface{}) error {
+	dConfig := &mapstructure.DecoderConfig{
```
lovromazgon, line 54: I found [this option](https://github.com/mitchellh/mapstructure/blob/bf980b35cac4dfd34e05254ee5aba086504c3f96/mapstructure.go#L264-L266) - thinking out loud if it would make sense to rename the tag to something more generic like `name`, `config` or maybe even just take the content of the `json` tag? 🤔

util_test.go
lovromazgon: Cool, I see the hook you added works for `[]T`, that's pretty neat! I thought it would only work for `[]string`.

util.go
maha-hajja: hmmm `name` wouldn't make sense when you use squash for example ``name:",squash"``, but I think `config` sounds good, donnow about `json` though

lovromazgon: I see now the linter is complaining because of `json:",squash"`, but now that I think about it I think it would make sense to squash embedded structs by default by setting `Squash` to `true` in `DecoderConfig`. Then you can embed a global config struct in your config as we do in [s3 for example](https://github.com/ConduitIO/conduit-connector-s3/blob/641f6222ec0d191802f5d00f0dff3b66c9dc55de/source/config.go#L35).

lovromazgon (approved):
LGTM! 🙌 Wasn't as easy as we initially thought, I'm glad we now provide this utility 😉

PR #44, opened 2022-12-12
adding paramgen

maha-hajja:
### Description

parse a Config struct into a map[string]sdk.Parameter and create a file with a function that returns the parameters

Fixes https://github.com/ConduitIO/conduit/issues/696

### Quick checks:

  update/change.


cmd/paramgen/internal/paramgen.go
```diff
@@ -0,0 +1,628 @@
...
+		prefix = p.formatFieldName(n.Name)
+	}
+	if prefix == "" {
+		// no prefix to attach
+		return params
+	}
+
+	prefixedParams := make(map[string]sdk.Parameter)
+	for k, v := range params {
+		prefixedParams[prefix+"."+k] = v
```
lovromazgon, line 446: Since we have separators in constants it would be nice to have the dot in a constant as well.

cmd/paramgen/internal/paramgen.go
```diff
@@ -0,0 +1,628 @@
...
+	tagGreaterThan1      = "gt"
+	tagGreaterThan2      = "greater-than"
+	tagInclusion         = "inclusion"
+	tagExclusion         = "exclusion"
+	pipeSeparator        = "|"
+)
+
+// ParseParameters parses the struct into a map of parameter, requires the folder path that has the struct, and the
+// struct name
+func ParseParameters(path string, name string) (map[string]sdk.Parameter, error) {
```
lovromazgon, line 52: Btw, seems like this is the only thing we need to expose, we could make everything else private (constants, `Module`, `ModuleError`).

cmd/paramgen/internal/paramgen.go
```diff
@@ -0,0 +1,628 @@
...
+	// can write normal go docs referencing the field name
+	desc := strings.ReplaceAll(f.Doc.Text(), fieldName, name)
+	desc = strings.ReplaceAll(desc, "\n", " ")
+	desc = strings.Trim(desc, " ")
+
+	return name, sdk.Parameter{
+		Default:     p.getTag(f.Tag, TagParamDefault),
+		Description: desc,
+		Validations: validations,
+		Required:    isRequired,
```
lovromazgon, line 509: This is a deprecated field, we shouldn't set it. The information will already be included in `Validations`.

cmd/paramgen/main.go
```diff
@@ -0,0 +1,152 @@
...
+	str = strings.ReplaceAll(str, "}}", "},\n}")
+	return str
+}
+
+func parseArgs() Args {
+	flags := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	var (
+		output     = flags.String("output", "conduit-config.go", "name of the output file")
+		structName = flags.String("struct", "Config", "the configuration struct name")
+		path       = flags.String("path", "/Users/mahahajja/repos/conduit-connector-sdk/cmd/paramgen/internal/testdata/tags", "directory path to the package that has the configuration struct")
```
lovromazgon, line 128: The default seems a bit specific, maybe we can make it a bit more generic for users that are not @maha-hajja? 🙃

cmd/paramgen/main.go
```diff
@@ -0,0 +1,152 @@
...
+	str = strings.ReplaceAll(str, "{", "{\n")
+	str = strings.ReplaceAll(str, "}}", "},\n}")
+	return str
+}
+
+func parseArgs() Args {
+	flags := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	var (
+		output     = flags.String("output", "conduit-config.go", "name of the output file")
+		structName = flags.String("struct", "Config", "the configuration struct name")
```
lovromazgon, line 127: I wouldn't have a default for this option. In fact, I would change this to be an argument instead of a flag and make it required for the user to provide it.

cmd/paramgen/main.go
```diff
@@ -0,0 +1,152 @@
...
+	str := fmt.Sprintf("%#v\n", params)
+	str = strings.ReplaceAll(str, "{", "{\n")
+	str = strings.ReplaceAll(str, "}}", "},\n}")
+	return str
+}
+
+func parseArgs() Args {
+	flags := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	var (
+		output     = flags.String("output", "conduit-config.go", "name of the output file")
```
lovromazgon, line 126: Hmm.. The default name seems strange. First it's the `-` in the filename, generally Go filenames use a `_` when necessary. But more importantly, I'd rather include "param" or "paramgen" somewhere in the name. Maybe `paramgen.go`? 🤔

cmd/paramgen/main.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	// convert the parameter map to a string
+	str := paramsToString(params)
+
+	processTemplate(str, args)
+}
+
+const (
+	outputDirName = "/paramgen"
```
lovromazgon, line 58: But why? 🤔 I think we should just output the file in the current folder. We would have to figure out the package name though (we already know it when parsing the files in `ParseParameters`).

cmd/paramgen/main.go
```diff
@@ -0,0 +1,152 @@
...
+	paramGenTmpl  = `// Code generated by ParamGen. DO NOT EDIT.
+// Source: github.com/conduitio/conduit-connector-sdk/cmd/paramgen
+
+// Package paramgen is a generated ParamGen package.
+package paramgen
+
+import sdk "github.com/conduitio/conduit-connector-sdk"
+
+func getParameters() map[string]sdk.Parameter {
+  return {{.}}
```
lovromazgon, line 68: Gofmt would not be happy with the two spaces, a tab should be used for indentation.

More importantly - the way we use the template right now doesn't bring us a lot of value. We are essentially just inserting a string into this template, this could easily be done with `fmt.Sprintf`. What I expected is that we would write a template that operates on `map[string]sdk.Parameter` and knows how to print that map as Go code. This would give us more control over the code that we output.

cmd/paramgen/main.go
```diff
@@ -0,0 +1,152 @@
...
+}
+`
+)
+
+// processTemplate creates and executes the paramgen template, then writes the output file under the directory ./paramgen.
+func processTemplate(str string, args Args) {
```
lovromazgon, line 74: I think this should be part of the `internal` package, `main` should only be an entry point without much logic apart from parsing flags and arguments. It would also be nice seeing some tests for this (tip: it will be easier to test if we return a string without writing it into a file).

cmd/paramgen/main.go
maha-hajja: lol I was testing locally and forgot to change it!

cmd/paramgen/internal/paramgen.go
maha-hajja: It would yes, and I ignored it at first, but it seemed weird in the output file that all the required fields are false, so I added it.
But yes I agree.

cmd/paramgen/internal/paramgen_test.go
```diff
@@ -0,0 +1,186 @@
...
+			name: "Config",
+			pkg:  "tags",
+			want: map[string]sdk.Parameter{
+				"innerConfig.my-name": {
+					Type:        sdk.ParameterTypeString,
+					Validations: []sdk.Validation{sdk.ValidationRequired{}},
+				},
+				"my-param": {
+					Type:        sdk.ParameterTypeInt,
+					Description: "my-param i am a parameter comment",
```
hariso, line 112: Tiny nitpick: maybe have a test with a line comment as well?

hariso (commented):
Great work!

cmd/paramgen/internal/paramgen.go
```diff
@@ -0,0 +1,628 @@
...
+	fset := token.NewFileSet()
+	filterTests := func(info fs.FileInfo) bool {
+		return !strings.HasSuffix(info.Name(), "_test.go")
+	}
+	pkg, err := parser.ParseDir(fset, path, filterTests, parser.ParseComments)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't parse directory %v: %w", path, err)
+	}
+	if len(pkg) != 1 {
+		return nil, errors.New("more than 1 package")
```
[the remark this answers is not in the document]
lovromazgon: Yes, If there's no source code in that folder. I'll add a check to make sure we produce a nicer error message in that case.

cmd/paramgen/internal/paramgen.go
```diff
@@ -0,0 +1,628 @@
...
+	panic("unreachable")
+}
+
+func findStruct(pkg *ast.Package, name string) (*ast.StructType, *ast.File, error) {
+	var structType *ast.StructType
+	var file *ast.File
+	for _, f := range pkg.Files {
+		ast.Inspect(f, func(n ast.Node) bool {
+			// Check if the node is a struct declaration
+			if typeSpec, ok := n.(*ast.TypeSpec); ok && typeSpec.Type != nil && typeSpec.Name.String() == name {
```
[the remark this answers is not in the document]
lovromazgon: You guessed correctly 🙂 A Go type can't contain spaces, so if `name` contains spaces it simply won't match any Go type.

cmd/paramgen/internal/paramgen.go
```diff
@@ -0,0 +1,628 @@
...
+	myParams, err := (*paramsParser)(p).parseStructType(structType)
+	if err != nil {
+		return nil, "", err
+	}
+
+	return myParams, pkgName, nil
+}
+
+// paramsParser groups functions that concern themselves with parsing parameters
+type paramsParser parameterParser
```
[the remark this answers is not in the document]
lovromazgon: I will get rid of `paramsParser` entirely.

cmd/paramgen/internal/testdata/basic/specs.go
```diff
@@ -0,0 +1,63 @@
...
+
+// SourceConfig this comment will be ignored.
+type SourceConfig struct {
+	GlobalConfig
+
+	// MyString my string description
+	MyString string
+	MyBool   bool
+
+	MyInt    int `validate:"lt=100, gt=0"`
```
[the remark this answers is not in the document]
lovromazgon: Hmm, it's not ideal. I'll see what I can do about it.

cmd/paramgen/internal/paramgen_test.go
lovromazgon: Good point 👍

hariso: My comment wasn't really clear, I meant comments which come in the same line as the code, e.g.
```go
myField int // here's a comment
```
If the above makes any difference that is.
But good that you mentioned block comments as well. :+1:

hariso (approved):
:rocket: :rocket:

cmd/paramgen/internal/paramgen_test.go
lovromazgon: I added support for line comments. They are only used if the doc comment on the field does not exist. So like this:
```
type Config struct {
	// Foo has a doc comment
	Foo int // line comment is ignored
	Bar // Bar only has a line comment so it's used
}
```

PR #46, opened 2022-12-13
Ignore validations for optional parameters

lovromazgon:
### Description

Fixes the bug where optional parameters are still validated when empty.

### Quick checks:

  update/change.

### DIFF
--- a/validation_test.go
+++ b/validation_test.go
@@ -235,7 +235,6 @@ func TestValidation_Param_Value(t *testing.T) {
 			params: map[string]Parameter{
 				"param1": {Validations: []Validation{
 					ValidationRequired{},
-					ValidationRequired{},
 				}},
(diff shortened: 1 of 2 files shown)


validation_test.go
```diff
@@ -235,6 +235,7 @@ func TestValidation_Param_Value(t *testing.T) {
 			params: map[string]Parameter{
 				"param1": {Validations: []Validation{
 					ValidationRequired{},
+					ValidationRequired{},
```
maha-hajja, line 238: required validation is duplicated here

maha-hajja (approved):

validation_test.go
lovromazgon: Yeah, because that param is _super important_ 😂

PR #51, opened 2023-01-06
fix unpopulated default values bug

### LINKED ISSUE #50: Bug: config default values are not populated
### Bug description

after the builtin validations are done in the SDK, then the custom validations that's located in the `Configure` method that the developer wrote is called.
However, the validator assigns the default values and validates them, but doesn't pass the new configuration map with the default values to the `impl.Configure` method. Which causes the default values to be lost.

### Steps to reproduce

1. create an S3 source without specifying the pollingPeriod parameter.
2. the connector will fail, because pollingPeriod value would be 0 and not "1s" > "1s" is the default value

### Version

v0.4.2

maha-hajja:
### Description

The fix is returning an updated map of Config that has the default values, and using that map when calling `impl.Configure()`

Fixes #50 

### Quick checks:

  update/change.

### DIFF
--- a/destination.go
+++ b/destination.go
@@ -89,12 +89,11 @@ type destinationPluginAdapter struct {
 func (a *destinationPluginAdapter) Configure(ctx context.Context, req cpluginv1.DestinationConfigureRequest) (cpluginv1.DestinationConfigureResponse, error) {
 	ctx = DestinationWithBatch{}.setBatchEnabled(ctx, false)
 
-	var multiErr error
 	// run builtin validations
-	multiErr = multierr.Append(multiErr, validator(a.impl.Parameters()).Validate(req.Config))
+	updatedCfg, multiErr := validator(a.impl.Parameters()).Validate(req.Config)
 	// run custom validations written by developer
-	multiErr = multierr.Append(multiErr, a.impl.Configure(ctx, req.Config))
-	multiErr = multierr.Append(multiErr, a.configureWriteStrategy(ctx, req.Config))
+	multiErr = multierr.Append(multiErr, a.impl.Configure(ctx, updatedCfg))
+	multiErr = multierr.Append(multiErr, a.configureWriteStrategy(ctx, updatedCfg))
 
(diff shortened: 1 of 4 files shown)


validation.go
```diff
@@ -210,7 +210,7 @@ func (v validator) Validate(config map[string]string) error {
 		}
 		multiErr = multierr.Append(multiErr, v.validateParamValue(pKey, cfg[pKey]))
 	}
-	return multiErr
+	return cfg, multiErr
```
hariso, line 213: Nitpick: At least to me, it doesn't feel quite intuitive that a Validate method also returns a fully populated config. IMHO it would be nice if we either rename it, or just call two methods. 

Side question: does the validator check the default values too?

hariso (approved):

hariso (approved):
:zap:

destination.go
```diff
@@ -89,12 +89,14 @@ type destinationPluginAdapter struct {
 func (a *destinationPluginAdapter) Configure(ctx context.Context, req cpluginv1.DestinationConfigureRequest) (cpluginv1.DestinationConfigureResponse, error) {
 	ctx = DestinationWithBatch{}.setBatchEnabled(ctx, false)
 
-	var multiErr error
+	v := validator(a.impl.Parameters())
+	// init config and apply default values
+	updatedCfg, multiErr := v.initConfig(req.Config)
```
lovromazgon, line 94: Now that `initConfig` is meant to be called outside of the validator we should probably make it public (i.e. rename to `InitConfig`).

maha-hajja: should I take it out the validator type too? so just a static method? or keep it

lovromazgon: I'd leave it on the validator rather than have a static function. The function needs the map of parameters and the validator already has access to them. If we feel like this should not be part of the validator I'd rather rename `validator` and give it a more generic name like `parameterUtil`. I personally can live with `validator` initializing the config though 😉

maha-hajja: done

lovromazgon (approved):

PR #53, opened 2023-01-10
Housekeeping: format paramgen readme

lovromazgon:
Just a reformat to make it display nicely.


hariso (approved):

maha-hajja:
I was thinking about adding a note to make sure $GOPATH/bin is in the PATH, is it too obvious?
something like mockgen https://github.com/golang/mock#installation

PR #55, opened 2023-01-25
Template Record Formatter

lovromazgon:
### Description

This adds a template record formatter. The user gets full control over the output format by using [go templates](https://pkg.go.dev/text/template). The value provided to the template is a [`sdk.Record`](https://github.com/ConduitIO/conduit-connector-sdk/blob/522125606be5ba8e4790b4300c9f577df21c2cbc/record.go#L80), so the template has access to all its fields (e.g. `.Position`, `.Key`, `.Metadata` and so on). We also inject all template functions provided by [sprig](http://masterminds.github.io/sprig/) to make it easier to write templates.

For example, running this pipeline syncs lines from file `example.in` to file `example.out` (without outputting the opencdc record and without base64 encoded payload):
```yaml
version: 1.1
pipelines:
  file-to-file:
    status: running
    connectors:
      example.in:
        type: source
        plugin: builtin:file
        settings:
          path: ./example.in
      example.out:
        type: destination
        plugin: file
        settings:
          path: ./example.out
          # use the template record formatter
          sdk.record.format: template
          # print only the payload, format it as a string (normally it's a byte slice)
          sdk.record.format.options: '{{ printf "%s" .Payload.After }}'
```

Fixes https://github.com/ConduitIO/conduit/issues/831.

### Quick checks:

  update/change.


hariso (approved):
LGTM! Two things only, both non-blocker:

1. If we can document this somewhere, to make it "visible" and known that we have this feature, it would be great.
2. I'm wondering what's the performance of this, especially in high volume pipelines. Performance is probably not a huge concern now, but for real-time pipelines with a lot of data it might become.

lovromazgon:
1. Definitely! 💯
2. If [these benchmarks](https://github.com/slinso/goTemplateBenchmark#full-featured-template-engines-1) are to be trusted, the `text/template` package is pretty fast, looks like it shouldn't create a bottleneck. Of course it depends on the complexity of the template itself, but I expect these to be relatively simple most of the time.

maha-hajja (approved):
lgtm

PR #58, opened 2023-02-07
Debezium raw data

lovromazgon:
### Description

This PR makes sure that records that contain actual raw data can be converted to debezium records by hoisting the raw data into a field. By default that field is named `opencdc.rawData` but it can be changed by setting `debezium.rawData.key`.

### Quick checks:

  update/change.

### DIFF
--- a/record_formatter_test.go
+++ b/record_formatter_test.go
@@ -53,11 +53,13 @@ func TestDebeziumConverter_Structured(t *testing.T) {
 
 	r := Record{
 		Position:  Position("foo"),
-		Operation: OperationCreate,
+		Operation: OperationUpdate,
 		Metadata:  Metadata{MetadataConduitSourcePluginName: "example"},
 		Key:       RawData("bar"),
 		Payload: Change{
-			Before: nil,
+			Before: StructuredData{
+				"bar": 123,
+			},
 			After: StructuredData{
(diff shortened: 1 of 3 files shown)


record_formatter_test.go
```diff
@@ -141,6 +143,95 @@ func TestDebeziumConverter(t *testing.T) {
...
+	is.NoErr(err)
+
+	r := Record{
+		Position:  Position("foo"),
+		Operation: OperationCreate,
+		Metadata:  Metadata{MetadataConduitSourcePluginName: "example"},
+		Key:       RawData("bar"),
+		Payload: Change{
+			Before: RawData("foo"),
+			After:  nil,
```
maha-hajja, line 158: should we maybe test both `Before` and `After` together?

maha-hajja (approved):
lgtm

record_formatter_test.go
hariso: Can you remind me, what happens if key is raw data too?

lovromazgon: The key doesn't make a difference, the debezium record only contains the payload. The kafka connector produces a kafka connect key if it sees that `sdk.record.format` is set to `debezium`.

hariso (approved):

PR #61, opened 2023-02-22
Paramgen: don't add prefix if nested field is not named

lovromazgon:
### Description

This fixes a mismatch between `paramgen` and `sdk.Util.ParseConfig`, where `ParseConfig` did not expect that embedded structs without a name would have a prefix (see [this test case](https://github.com/ConduitIO/conduit-connector-sdk/blob/29058268e9933edc3f99fcb62f01015a4c778fbd/util_test.go#L57)). Now `paramgen` also does not add that prefix to generated parameters.

### Quick checks:

  update/change.


maha-hajja (approved):
lgtm

PR #77, opened 2023-05-26
Allow SDK to detach from stuck connectors

lovromazgon:
### Description

This change ensures the following:

- When `Teardown` is called without `Stop`, the context in `Open` and `Read` will still get cancelled. This could happen when the shutdown is not graceful, but rather it was triggered because an error happened in Conduit. In that case Conduit calls `Teardown` directly without draining records.
- If the connector gets stuck and not responding to a stopped context (i.e. keeps on reading records or waiting for new records) the SDK is now able to detach from it. In `Stop` it will wait 1 minute for the connector to stop reading new records, and another 1 minute in `Teardown`. After that it kills the tomb which stops the stream forcefully.

**Important** - even though this should add better resilience to recover from stuck connectors, the timeouts should be revisited or made configurable, as 1 minute could sometimes be legitimately too short for a connector to drain the cached records. Imagine the case when the backpressure from a slow destination is preventing the connector from draining the records - we will detach from it, even though it would eventually gracefully stop. Maybe the decision should be left to the user to trigger a force stop instead of having these timeouts. Let's discuss the expected behavior.

### Quick checks:

  update/change.

### DIFF
--- a/source.go
+++ b/source.go
@@ -30,6 +30,12 @@ import (
 	"gopkg.in/tomb.v2"
 )
 
+var (
+	// TODO make the timeout configurable (https://github.com/ConduitIO/conduit/issues/183)
+	stopTimeout     = time.Minute
+	teardownTimeout = time.Minute
+)
+
 // Source fetches records from 3rd party resources and sends them to Conduit.
 // All implementations must embed UnimplementedSource for forward compatibility.
 type Source interface {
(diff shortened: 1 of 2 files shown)


source_test.go
```diff
@@ -199,6 +200,66 @@ func TestSourcePluginAdapter_Run(t *testing.T) {
...
 
+func TestSourcePluginAdapter_Run_Stuck(t *testing.T) {
+	is := is.New(t)
+	ctrl := gomock.NewController(t)
+	src := NewMockSource(ctrl)
+
+	teardownTimeout = time.Second
+	stopTimeout = time.Second
+	defer func() {
+		teardownTimeout = time.Minute // reset
```
maha-hajja, line 211: this feels a bit risky, what if something happened and the defer function didn't run? is that a possibility?

maha-hajja (commented):
looks good! and I think 1 minute is a good start until we make them configurable

source.go
```diff
@@ -240,30 +246,45 @@ func (a *sourcePluginAdapter) runAck(ctx context.Context, stream cpluginv1.Sourc
...
-	// TODO timeout for badly written connectors
-	<-a.readDone // wait for read to actually stop running
+	// wait for read to actually stop running with a timeout, in case the
+	// connector gets stuck
+	waitCtx, cancel := context.WithTimeout(ctx, stopTimeout)
+	defer cancel()
+
+	err := a.waitForClose(waitCtx, a.readDone)
+	if err != nil {
+		Logger(ctx).Warn().Err(err).Msg("failed to wait for Read to stop running")
```
hariso, line 261: Nitpick: unless `waitForClose` already does it, it would be good to mention the timeout value in the logs.

source_test.go
hariso: Maybe it's better to make the timeouts "partially configurable", i.e. make them fields in `sourcePluginAdapter`. That would make them more usable in tests, less risky, and that code can be re-used when we make it "completely configurable".

hariso (approved):
:zap:

maha-hajja:
since @lovromazgon  is off for a while (and I need these changes), I'm gonna merge this PR for now, and he can create another one addressing the reviews later.
@hariso

PR #80, opened 2023-07-10
Destination record batching

### LINKED ISSUE #74: Feature: Destination middleware for batching
### Feature description

We should implement and enable the Destination middleware for [batching records](https://github.com/ConduitIO/conduit-connector-sdk/blob/e00e0c375000e48ff27ccce0c410988cec085b41/destination_middleware.go#L38).

lovromazgon:
### Description

Implements record batching in destinations. Batching can be turned on using the connector config options `sdk.batch.size` and `sdk.batch.delay`, e.g.:

```
sdk.batch.size: 10000 # trigger batch when 10000 records are collected
sdk.batch.delay: 2s   # trigger batch after 2 seconds even if it didn't reach the full size
```

Fixes #74 

### Quick checks:

  update/change.

### DIFF
--- a/kafkaconnect/reflect.go
+++ b/kafkaconnect/reflect.go
@@ -129,6 +129,7 @@ func reflectInternal(v reflect.Value, t reflect.Type) *Schema {
 		for _, kv := range v.MapKeys() {
 			keyValue = kv
 			valValue = v.MapIndex(kv)
+			break
 		}
 		return &Schema{
 			Type:   TypeMap,
--- a/destination.go
+++ b/destination.go
@@ -19,7 +19,6 @@ package sdk
 import (
 	"bytes"
 	"context"
-	"errors"
 	"fmt"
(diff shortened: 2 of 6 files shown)


kafkaconnect/reflect.go
```diff
@@ -129,6 +129,7 @@ func reflectInternal(v reflect.Value, t reflect.Type) *Schema {
 		for _, kv := range v.MapKeys() {
 			keyValue = kv
 			valValue = v.MapIndex(kv)
+			break
```
hariso, line 132: Intended change?

hariso (approved):

kafkaconnect/reflect.go
lovromazgon: Yes, we just need a value from the map, so we don't need to range over all keys, the first value should work just fine.

PR #87, opened 2023-10-05
Fix acks sent to Conduit on error in batch mode

lovromazgon:
### Description

I spotted an issue with acks returned from the connector when used in batch mode. The connector didn't return all acks, while it should actually return acks and include the error. This PR fixes it and adds two tests for testing batch writes.

### Quick checks:

  update/change.


destination.go
```diff
@@ -382,16 +382,20 @@ func (w *writeStrategyBatch) writeBatch(batch []writeBatchItem) error {
 	}
 
 	var (
-		firstErr error
-		errOnce  bool
+		ackRespose error
```
hariso, line 382: Typo (missing 'n' in response)

hariso (approved):

hariso:
Great catch!

PR #96, opened 2023-10-31
Benchmark for source connectors

lovromazgon:
### Description

This PR adds the function `BenchmarkSource` which can be used to benchmark connectors. It originated in the [Kafka connector](https://github.com/ConduitIO/conduit-connector-kafka/blob/ffbf30b6174c3db27ded2d4748170a0181fe05f0/cmd/th/main.go) where this was a utility to test the throughput of the connector. It was generalized and made into a Go benchmark.

Running the benchmark for the Kafka connector looks like this:
```
go test -bench=BenchmarkSource -run=^$
goos: darwin
goarch: amd64
pkg: github.com/conduitio/conduit-connector-kafka
cpu: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
BenchmarkSource-16    	 1277617	   1074857 acks/s	         0.0000726 configure	         0.0000005 firstAck	         3.092 firstRead	   1074951 reads/s	         0.001811 open	         0.0000414 stop	         0.003079 teardown
PASS
ok  	github.com/conduitio/conduit-connector-kafka	19.681s
```

It prints out the duration in seconds for each part of the lifecycle (`configure`, `open`, `stop`, `teardown`), the duration in seconds for the first ack and first read (`firstAck`, `firstRead`) and most importantly the throughput of reads and acks (`reads/s`, `acks/s`).


benchmark.go
```diff
@@ -0,0 +1,175 @@
...
+		err := bm.source.Ack(ctx, rec.Position)
+		if err != nil {
+			b.Fatal("Ack: ", err)
+		}
+	})
+
+	count := 0
+	bm.measure(&bm.allAcks, func() {
+		for rec := range c {
+			count++
```
simonl2002, line 147: What's count for?

simonl2002 (approved):
lgtm, though had one question.

benchmark.go
lovromazgon: Hmm 🤔 I wonder as well

PR #115, opened 2024-04-07
[WIP] Utility for generating readme from specifications

lovromazgon:
### Description

This PR contains an experimental utility `readmegen` that uses the connector specification to generate the connector readme. The specification contains a lot of useful information about how to configure the connector, but it's not enough to generate the full readme, as there might be intricacies that the user wants to add. Still, this could be a useful tool to generate the configuration info and keep it up to date.

Why WIP? We need to figure out how this could fit in the development flow of implementing a connector. Ideally we should be able to trigger a make target that outputs the updated readme, without the need to manually add more stuff. This would allow us to run it as part of a CI action that checks if generated files are up to date ([example](https://github.com/ConduitIO/conduit-commons/blob/main/.github/workflows/validate-generated-files.yml)), this way we'd ensure that the readme contains the newest information after an update of the SDK or after we change some configuration parameters.

### DIFF
--- a/cmd/readmegen/main.go
+++ b/cmd/readmegen/main.go
@@ -1,0 +1,164 @@
+// Copyright © 2024 Meroxa, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"text/template"
+
+	"github.com/conduitio/conduit-connector-sdk/cmd/readmegen/util"
+	"golang.org/x/mod/modfile"
+)
+
+var (
+	pkg               = flag.String("package", "", "The full package import path for the connector. By default, readmegen will try to detect the package import path based on the go.mod file it finds in the directory or any parent directory.")
+	debugIntermediary = flag.Bool("debug-intermediary", false, "Print the intermediary generated program to stdout instead of writing it to a file")
+	yaml              = flag.Bool("yaml", false, "Generate parameters in YAML format (default is a HTML table)")
+)
+
+func main() {
+	flag.Parse()
+
+	if *pkg == "" {
+		var err error
+		*pkg, err = detectPackage()
+		if err != nil {
+			fmt.Fprintln(os.Stderr, err)
+			os.Exit(1)
+		}
+	}
+
+	err := generate()
+	if err != nil {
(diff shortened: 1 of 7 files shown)


conn-sdk-cli/main.go
```diff
@@ -0,0 +1,72 @@
...
+	"fmt"
+	"os"
+
+	"github.com/conduitio/conduit-connector-sdk/conn-sdk-cli/readmegen"
+	"github.com/conduitio/conduit-connector-sdk/conn-sdk-cli/specgen"
+	"github.com/spf13/cobra"
+)
+
+func main() {
+	cmdRoot := &cobra.Command{
```
hariso, line 27: WDYT about trying ecdysis here too?

conn-sdk-cli/readmegen/preprocess.go
```diff
@@ -0,0 +1,126 @@
...
+		"source.parameters.yaml":       `{{ template "parameters.yaml" args "specification" .specification "parameters" .specification.source.parameters }}`,
+		"source.parameters.table":      `{{ template "parameters.table" args "specification" .specification "parameters" .specification.source.parameters }}`,
+		"destination.parameters.yaml":  `{{ template "parameters.yaml" args "specification" .specification "parameters" .specification.destination.parameters }}`,
+		"destination.parameters.table": `{{ template "parameters.table" args "specification" .specification "parameters" .specification.destination.parameters }}`,
+	}
+	errReadmegenCommentNotFound = errors.New("readmegen open tag not found")
+)
+
+// Preprocess takes the contents of a readme file and preprocesses it by replacing
+// readmegen tags with the corresponding values. A readmegen tag is a HTML comment
```
hariso, line 52: "replacing readmegen tags with the corresponding values" makes me think that Preprocess produces a string with values, but it actually transforms the readme with tags into a readme Go template?

conn-sdk-cli/main.go
lovromazgon: We could of course do that, I'm not sure if it's worth it now though. I went with vanilla cobra here as I needed something simple for 2 commands only, while ecdysis shines in organizing commands in bigger CLIs IMO.

conn-sdk-cli/readmegen/generate_test.go
```diff
@@ -0,0 +1,20 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package readmegen
+
+import "testing"
+
+func TestGenerate(t *testing.T) {
```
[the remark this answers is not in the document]
lovromazgon: Oh, I meant to implement an end-to-end test here.

conn-sdk-cli/readmegen/preprocess.go
lovromazgon: Rephrased `values` to `Go templates` 👍

conn-sdk-cli/main.go
hariso: Makes sense :+1: .

conn-sdk-cli/readmegen/generate_test.go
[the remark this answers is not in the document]
lovromazgon: I added a test and noticed that zero defaults weren't formatted nicely, also fixed that in the last commit. Good thing we write tests 😄

hariso (approved):

PR #116, opened 2024-04-08
Add introduced metadata constants, use `go.uber.org/mock`

lovromazgon:
### Description

Adds constants introduced in https://github.com/ConduitIO/conduit-connector-protocol/pull/73. Also replaces the deprecated package `github.com/golang/mock/gomock` with `go.uber.org/mock/gomock`.

Closes https://github.com/ConduitIO/conduit/issues/1472

### Quick checks:

  update/change.
- [X] I have written unit tests. (existing ones)


go.mod
```diff
@@ -1,47 +1,53 @@
 module github.com/conduitio/conduit-connector-sdk
 
-go 1.21
+go 1.21.1
```
raulb, line 3: @lovromazgon could we also bump to 1.22.2 here?

raulb (approved):

go.mod
lovromazgon: The SDK is imported by connectors, which means that the next time we update the SDK in the connectors (or when Dependabot does that for us) we'll need to bump the Go version in those as well. In other words, this change potentially creates a lot of housekeeping work, I wouldn't want to do it unless there's a good reason.

raulb: @lovromazgon makes sense. It could be a combined effort to do all of those at once. 👍

lovromazgon: Actually, I realized my answer is outdated since the introduction of [toolchains in Go 1.21](https://go.dev/doc/toolchain). The TL;DR is that this line advertises the _minimum_ Go version as `1.21.1`, so any dependency needs to use _at least_ version `1.21.1`. The toolchain line 2 lines below denotes that the preferred toolchain for this module is `go1.22.0`. I'm not sure if this is reflected in dependencies as well, but regardless, it doesn't prevent connectors from upgrading their Go version as `1.22.2` (or any newer version), it just specifies the minimum.

PR #118, opened 2024-04-15
Include package names in error

hariso:
### Quick checks:

  update/change.

### DIFF
--- a/cmd/paramgen/internal/paramgen.go
+++ b/cmd/paramgen/internal/paramgen.go
@@ -20,7 +20,6 @@ import (
 	"go/ast"
 	"go/parser"
 	"go/token"
-	"golang.org/x/exp/maps"
 	"io/fs"
 	"os/exec"
 	"reflect"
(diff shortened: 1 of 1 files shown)


cmd/paramgen/internal/paramgen.go
```diff
@@ -123,7 +124,7 @@ func parsePackage(path string) (*ast.Package, error) {
 		return nil, fmt.Errorf("no source-code package in directory %s", path)
 	}
 	if len(pkgs) > 1 {
-		return nil, fmt.Errorf("multiple packages in directory %s", path)
+		return nil, fmt.Errorf("multiple packages [%v] in directory %s", maps.Keys(pkgs), path)
```
lovromazgon, line 127: We should remove the brackets, otherwise this will print `[[key1 key2]]`.
```suggestion
		return nil, fmt.Errorf("multiple packages %v in directory %s", maps.Keys(pkgs), path)
```

lovromazgon (approved):
LGTM, a single nitpick. Do you mind adding the same change in conduit-commons please? We should try to keep them in sync until we remove this paramgen and migrate to conduit-commons.

PR #130, opened 2024-05-14
Use cpluginv2 and conduit commons

lovromazgon:
### Description

The PR updates the connector protocol to the latest code in `main` (we need to bump the version before releasing).

Additionally, it removes the internal `Record` type and everything related to it (structured data, raw data, metadata, serializers etc.) in favor of the `opencdc.Record` type in conduit-commons. It also removes `Validation` in favor of `config.Validation` in conduit-commons.

⚠️  This creates a breaking change, connectors updating to this change will need to be changed.

~Depends on https://github.com/ConduitIO/conduit-connector-protocol/pull/90.~ Merged. 

Part of https://github.com/ConduitIO/conduit/issues/1433.

### Quick checks:

- [X] I have followed the [Code Guidelines](https://github.com/ConduitIO/conduit/blob/main/docs/code_guidelines.md).
- [X] There is no other [pull request](https://github.com/ConduitIO/conduit-connector-sdk/pulls) for the same
  update/change.
- [X] I have written unit tests.
- [X] I have made sure that the PR is of reasonable size and can be easily reviewed.

PR #133, opened 2024-05-15
Use conduit commons utils

lovromazgon:
### Description

This PR removes internal utilities like `cchan` and `csync` in favor of similar utilities in conduit-commons. It also removes the automatic config validation and applying of defaults in `Configure`, these calls were now moved to the utility function `ParseConfig`, which is already called by most, if not all our connectors. This will also make it possible to write tests for configuration types and make parsing behaves as expected.

Depends on #130.

### Quick checks:

  update/change.


hariso (approved):
:rocket:

PR #150, opened 2024-07-09
Allow insecure gRPC for connector utils

hariso:
### Description

Please include a summary of the change and what type of change it is (new feature, bug fix, refactoring, documentation).
Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

### Quick checks:

  update/change.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,10 +1,12 @@
 module github.com/conduitio/conduit-connector-sdk
 
-go 1.22.3
+go 1.22.4
+
+toolchain go1.22.5
 
 require (
 	github.com/Masterminds/sprig/v3 v3.2.3
-	github.com/conduitio/conduit-commons v0.2.1-0.20240701165122-3948a38f3667
+	github.com/conduitio/conduit-commons v0.2.1-0.20240709142247-d973cba9694c
 	github.com/conduitio/conduit-connector-protocol v0.6.1-0.20240705154009-b938cfa7f251
 	github.com/goccy/go-json v0.10.3
 	github.com/golangci/golangci-lint v1.59.1
@@ -190,6 +192,7 @@ require (
 	github.com/timonwong/loggercheck v0.9.4 // indirect
 	github.com/tomarrell/wrapcheck/v2 v2.8.3 // indirect
 	github.com/tommy-muehle/go-mnd/v2 v2.5.1 // indirect
+	github.com/twmb/go-cache v1.2.1 // indirect
 	github.com/ultraware/funlen v0.1.0 // indirect
 	github.com/ultraware/whitespace v0.1.1 // indirect
 	github.com/uudashr/gocognit v1.1.2 // indirect
@@ -220,3 +223,5 @@ require (
 	mvdan.cc/gofumpt v0.6.0 // indirect
 	mvdan.cc/unparam v0.0.0-20240528143540-8a5130ca722f // indirect
 )
+
+replace github.com/conduitio/conduit-connector-protocol => ../conduit-connector-protocol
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -220,3 +223,5 @@ require (
 	mvdan.cc/gofumpt v0.6.0 // indirect
 	mvdan.cc/unparam v0.0.0-20240528143540-8a5130ca722f // indirect
 )
+
+replace github.com/conduitio/conduit-connector-protocol => ../conduit-connector-protocol
```
hariso, line 227: Reminder: remove once approved.

go.mod
```diff
@@ -1,11 +1,13 @@
...
+go 1.22.4
+
+toolchain go1.22.5
 
 require (
 	github.com/Masterminds/sprig/v3 v3.2.3
-	github.com/conduitio/conduit-commons v0.2.1-0.20240701165122-3948a38f3667
-	github.com/conduitio/conduit-connector-protocol v0.6.1-0.20240705154009-b938cfa7f251
+	github.com/conduitio/conduit-commons v0.2.1-0.20240709142247-d973cba9694c
+	github.com/conduitio/conduit-connector-protocol v0.6.1-0.20240711112538-4e32f3a34aa3
```
hariso, line 10: Reminder: update this dep once the protocol PR is merged.

serve.go
```diff
@@ -58,29 +58,59 @@ func serve(c Connector) error {
...
+		ConnectorID: "",
+		LogLevel:    0,
+	}, nil
+}
+
+// connectorUtilsAddress returns the address and token to be used for the connector utilities service.
+// The values are fetched from environment variables provided by conduit-connector-protocol.
+// The function returns an error if the environment variables are not specified or empty.
+func connectorUtilsAddress() (string, error) {
+	target := os.Getenv(pconduit.EnvConduitConnectorUtilitiesGRPCTarget)
```
hariso, line 101: WDYT about renaming this env. var to `EnvConduitConnectorUtilitiesAddress`? Also maybe dropping `Conduit` from the prefix (the package is `pconduit`)?

lovromazgon: We talked about this briefly with Raul, I can tell you the reasoning behind the name. First, the constant name corresponds to the value of the environment variable prefixed with `Env`, which is just capitalized and in snake case (`CONDUIT_CONNECTOR_UTILITIES_GRPC_TARGET`). As for the name itself:
- `CONDUIT` - prefix to prevent clashes with any other variables
- `CONNECTOR_UTILITIES` - the terminology we use for these gRPC services
- `GRPC_TARGET` - comes from [`grpc.NewClient`](https://pkg.go.dev/google.golang.org/grpc#NewClient) where the first parameter is named `target`

I know it's verbose, but I feel like it's important to be explicit and descriptive when it comes to environment variables.

internal/context.go
```diff
@@ -0,0 +1,69 @@
...
+		return logLevel.(string) //nolint:forcetypeassert // only this package can set the value, it has to be a string
+	}
+
+	return ""
+}
+
+func Enrich(ctx context.Context, cfg pconnector.PluginConfig) context.Context {
+	ctx = pconduit.ContextWithConnectorToken(ctx, cfg.Token)
+	ctx = ContextWithConnectorID(ctx, cfg.ConnectorID)
+	ctx = ContextWithLogLevel(ctx, cfg.LogLevel)
```
lovromazgon, line 66: This is a high nitpick, but here it goes - I'd reverse the order in which we attach these values. Why? The context is essentially a stack, so fetching a value from the context means we have to pop values from the stack until we get to the one we are looking for. We know that the token will be used in calls to connector utils, so I'd keep it on top of the stack. I also expect the connector ID to be used more often than the log level, that's why I'd just reverse the order.

lovromazgon (approved):
LGTM, left a single comment 👍

internal/context.go
hariso: Ah, when debugging I noticed that contexts are wrapped, but I didn't immediately think it might have an impact on getting values. Thanks!

PR #152, opened 2024-07-15
Source middleware for extracting schema

### LINKED ISSUE #137: [Schemas] Add source middleware
### Feature description

The following functions are needed:

- auto-generate a schema
- register a schema in Conduit
- attach the schema metadata
- encode the payload

lovromazgon:
### Description

Adds a source middleware that extracts a schema from a record containing structured data, registers the schema in the schema registry, and encodes the value with the extracted schema. At the end, the schema subject and version are attached to the record metadata. This way we can retain all types end-to-end. The middleware is added as a default source middleware.

Additionally, this PR introduces `SourceMiddlewareOption` that can be passed to `DefaultSourceMiddleware` and can be used to configure the default source middleware with different defaults.

Fixes #137 

### Quick checks:

  update/change.


source_middleware.go
```diff
@@ -23,7 +37,13 @@ type SourceMiddleware interface {
 // all sources unless there's a good reason not to.
 func DefaultSourceMiddleware() []SourceMiddleware {
 	return []SourceMiddleware{
-		// TODO add default middleware (e.g. rate limit)
+		SourceWithSchema{
+			DefaultSchemaFormat:   schema.TypeAvro.String(),
```
hariso, line 41: Unrelated to this PR (since I see the same pattern in other middleware): how about we name the fields without the prefix `Default`? When I'm reading a middleware's code, the prefix `Default` is a bit confusing. Here we see from the context that it's about default values. And also when we use it in the form `foo.defaults.SchemaFormat` it's also clear.

source_middleware.go
```diff
@@ -34,3 +54,283 @@ func SourceWithMiddleware(d Source, middleware ...SourceMiddleware) Source {
...
+	configSourceSchemaPayloadSubject = "sdk.schema.payload.subject"
+	configSourceSchemaKeyEncode      = "sdk.schema.key.encode"
+	configSourceSchemaKeySubject     = "sdk.schema.key.subject"
+)
+
+// SourceWithSchema is a middleware that extracts and encodes the record payload
+// and key with a schema. The schema is extracted from the record data for each
+// record produced by the source. The schema is registered with the schema
+// service and the schema subject is attached to the record metadata.
+type SourceWithSchema struct {
```
hariso, line 72: Nit: maybe name the middleware so that it reflects that the schema is automatically extracted/registered? The middleware is doing a lot, which makes it a bit more difficult to find a good name.:)

lovromazgon: That's a valid concern, I was also thinking about what to call this middleware but didn't spend too much time figuring out a better suitable name. Maybe `SourceWithAutoSchemaExtraction`? 🤔

source_middleware.go
lovromazgon: I changed the field names, extracted the config into a separate type and added a way to configure the default middleware. This will make it easier to tweak the default middleware, previously it had to be done [awkwardly like this](https://github.com/conduitio-labs/conduit-connector-snowflake/blob/671448d3cb3718649b01a39447666cb6dac3c924/destination/destination.go#L50-L52). Let me know what you think, if we want to go with this I'll adjust the destination middleware to follow the same pattern.

source_middleware.go
hariso: To me at least, `SourceWithSchemaExtraction` sounds good enough, it kinda hints that the extraction is being done automatically.

source_middleware.go
```diff
@@ -14,17 +14,42 @@
...
+// SourceMiddlewareOption can be used to change the behavior of the default source
+// middleware created with DefaultSourceMiddleware.
+type SourceMiddlewareOption func(SourceMiddleware)
+
 // DefaultSourceMiddleware returns a slice of middleware that should be added to
 // all sources unless there's a good reason not to.
-func DefaultSourceMiddleware() []SourceMiddleware {
-	return []SourceMiddleware{
-		// TODO add default middleware (e.g. rate limit)
+func DefaultSourceMiddleware(opts ...SourceMiddlewareOption) []SourceMiddleware {
```
hariso, line 42: This makes it possible to configure default middleware. How about the case where we want to include only some of the middleware with `sdk.SourceWithMiddleware(&fooconnector.Source{}, sdk.BarMiddleware{})`?

source_middleware.go
```diff
@@ -14,17 +14,42 @@
...
+	middleware := []SourceMiddleware{
+		&SourceWithSchema{},
```
hariso, line 44: WDYT about assigning default values here? I think it would make it easier to discover it (since at least I would visit this method first).

source_middleware.go
```diff
@@ -34,3 +59,350 @@ func SourceWithMiddleware(d Source, middleware ...SourceMiddleware) Source {
...
+	if val, ok := config[configSourceSchemaKeyEncode]; ok {
+		encodeKey, err = strconv.ParseBool(val)
+		if err != nil {
+			return fmt.Errorf("invalid %s: failed to parse boolean: %w", configSourceSchemaKeyEncode, err)
+		}
+	}
+	if encodeKey {
+		// TODO: when adding schema context support, set DefaultKeySubject
+		//  to "key" and let the schema service attach the prefix / context.
+		s.keySubject = connectorID + ".key"
```
hariso, line 253: We're going to this when handling name conflicts, no?

source_middleware.go
lovromazgon: In that case you can set the defaults directly in the middleware struct, no need for options in that case:
```
sdk.SourceWithMiddleware(
    &fooconnector.Source{},
    sdk.BarMiddleware{
        Config: sdk.BarMiddlewareConfig{
            ParamFoo: "foo",
            ParamBar: 123,
        },
    },
)       
```

In other words, the options only play a role if you want to change the config of the default middleware, but retain the full list of the default middleware and benefit from new middleware if it gets added in future releases.

Alternatively, we could rename the options to `DefaultSourceMiddlewareOption`..?

hariso: Ah, right, good old struct init, no need for anything more "clever" than that!

As for `DefaultSourceMiddlewareOption`, it sounds good to me, but I'm strongly on this or either side.:)

source_middleware.go
lovromazgon: That's what I had before, but then I changed it, otherwise the defaults would be hardcoded here. If someone wanted to use this middleware with the same default configuration, they would have to copy over this part, and if we change the defaults, the copied code won't reflect the changes. So why I prefer this approach, is that the empty value already reflects the default behavior.

source_middleware.go
lovromazgon: Yeah, at that point this parameter can get a simple default value `"key"` and we will let the other parameter choose if a connector ID (or another prefix) gets added to make it unique and prevent name conflicts.

source_middleware.go
```diff
@@ -34,3 +59,350 @@ func SourceWithMiddleware(d Source, middleware ...SourceMiddleware) Source {
...
+	configSourceSchemaKeyEncode      = "sdk.schema.key.encode"
+	configSourceSchemaKeySubject     = "sdk.schema.key.subject"
+)
+
+// WithSourceWithSchemaConfig returns a SourceMiddlewareOption that sets the
+// default configuration for the SourceWithSchema middleware.
+func WithSourceWithSchemaConfig(cfg SourceWithSchemaConfig) SourceMiddlewareOption {
+	return func(m SourceMiddleware) {
+		if s, ok := m.(*SourceWithSchema); ok {
+			s.Config = cfg
```
hariso, line 78: Having all the config in one struct makes it convenient because we need to provide only one `SourceMiddlewareOption`. Thinking out loud: do we want to make it possible to change only individual config fields (without having to copy paste a default config, and then modifying only 1 or 2 fields)? It would mean many more options, which isn't that nice.

Another option is maybe to have a default config exposed? Then a developer could simply take that, override some values and done.

source_middleware.go
hariso: So maybe we can just remove the connector ID from this code right away?

lovromazgon: But then this code won't behave as it should 😕

hariso: Right... I can then change this in the name conflicts branches once everything is merged.

source_middleware.go
lovromazgon: That's another reason why I went with having the empty value reflect the default behavior, because then we can get away with a single option to pass in the config and you can only set the fields you care about, all the others will still retain the default behavior.

hariso: Gotcha, I re-read the `Wrap()` method just now and realized that. It would be nice to call it out in a comment or something, it's making it nice to customize a middleware!

hariso (approved):
:1st_place_medal:

source_middleware.go
lovromazgon: Ok seems like the pattern is set, in the next PR that introduces the destination middleware I'll apply it there as well.

source_middleware.go
lovromazgon: I'll leave it as it is then, to not make it more verbose than it has to be 🙃

source_middleware.go
lovromazgon: Ok, added a few comments in the config and `Wrap` to make this clearer 👍

PR #154, opened 2024-07-24
Destination middleware options

lovromazgon:
### Description

As mentioned in https://github.com/ConduitIO/conduit-connector-sdk/pull/152#discussion_r1684477189, this PR adds destination middleware options to match the source middleware options introduced in that PR.

### Quick checks:

  update/change.
- [X] I have written unit tests. (existing ones succeed, functionality did not change)

### DIFF
--- a/destination_middleware.go
+++ b/destination_middleware.go
@@ -31,14 +31,32 @@ type DestinationMiddleware interface {
 	Wrap(Destination) Destination
 }
 
+// DestinationMiddlewareOption can be used to change the behavior of the default
+// destination middleware created with DefaultDestinationMiddleware.
+type DestinationMiddlewareOption func(DestinationMiddleware)
+
+// Available destination middleware options.
+var (
+	_ DestinationMiddlewareOption = DestinationWithBatchConfig{}.Apply
+	_ DestinationMiddlewareOption = DestinationWithRecordFormatConfig{}.Apply
+	_ DestinationMiddlewareOption = DestinationWithRateLimitConfig{}.Apply
+)
+
 // DefaultDestinationMiddleware returns a slice of middleware that should be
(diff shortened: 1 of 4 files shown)


destination_middleware.go
```diff
@@ -58,6 +76,47 @@ const (
...
 
+// DestinationWithBatchConfig is the configuration for the
+// DestinationWithBatch middleware. Fields set to their zero value are
+// ignored and will be set to the default value.
+type DestinationWithBatchConfig struct {
+	// BatchSize is the default value for the batch size.
+	BatchSize int
+	// BatchDelay is the default value for the batch delay.
+	BatchDelay time.Duration
+}
```
hariso, line 87: WDYT about setting the default values through Wrap(), as is done in the source middleware?

hariso (approved):

destination_middleware.go
lovromazgon: Setting the default values through `Wrap` is ok with me. The question is _if_ we should set the defaults for batches, and if we do, what would be sane defaults? 😉 I think this might be a conversation for a discussion, as the implementation is trivial, we just need to consider what impact this could have on pipelines and what are sane defaults.

hariso: I was assuming that we are somewhere along way setting the batch size to 1. But if it works in a different, as you said,  no need to go into details here. :+1:

lovromazgon: I actually went and refactored this middleware a bit, so that it makes more sense and is more aligned with the other middleware, but didn't change the behavior or defaults. The zero values stayed the default, batching only kicks in if you set the batch size to something bigger than 1 or the delay to something bigger than 0.

hariso (approved):
:zap:

PR #155, opened 2024-07-27
Destination middleware for decoding data with attached schema

### LINKED ISSUE #138: [Schemas] Add destination middleware
### Feature description

Based on the metadata, the destination middleware fetches the schema and decodes the payload back into structured data.

lovromazgon:
### Description

This PR adds `DestinationWithSchemaExtraction` middleware, which is the counterpart to `SourceWithSchemaExtraction` introduced in https://github.com/ConduitIO/conduit-connector-sdk/pull/152. It checks the record metadata for an attached schema subject and version, fetches the schema and decodes the raw payload into a structured one using the obtained schema.

A few additional changes:
- Functions `DestinationWithMiddleware` and `SourceWithMiddleware` now apply middleware in the reverse order, so the order of calls between the middleware is in the same order as supplied by the caller.
- Parameters that end with `*.enable` changed to `*.enabled`.
- Rename a few tests.

Fixes #138

### Quick checks:

  update/change.

### DIFF
--- a/destination_middleware.go
+++ b/destination_middleware.go
@@ -16,13 +16,17 @@ package sdk
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sort"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/conduitio/conduit-commons/config"
 	"github.com/conduitio/conduit-commons/opencdc"
+	"github.com/conduitio/conduit-commons/schema"
+	sdkSchema "github.com/conduitio/conduit-connector-sdk/schema"
 	"golang.org/x/time/rate"
 )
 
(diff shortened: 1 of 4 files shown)


destination_middleware.go
```diff
@@ -498,3 +505,239 @@ func (d *destinationWithRecordFormat) Write(ctx context.Context, recs []opencdc.
...
+		// log warning once, to avoid spamming the logs
+		d.keyWarnOnce.Do(func() {
+			Logger(ctx).Warn().Msgf(`record does not have an attached schema for the key, consider disabling the destination schema key decoding using "%s: false"`, configDestinationWithSchemaExtractionKeyEnabled)
+		})
+		return nil
+	case errSubject != nil && !errors.Is(errSubject, opencdc.ErrMetadataFieldNotFound):
+		return fmt.Errorf("failed to get key schema subject from metadata: %w", errSubject)
+	case errVersion != nil && !errors.Is(errVersion, opencdc.ErrMetadataFieldNotFound):
+		return fmt.Errorf("failed to get key schema version from metadata: %w", errVersion)
+	}
```
hariso, line 666: WDYT about handling the case where one of the metadata fields is present and the other isn't? It should happen under normal circumstances but a bug or a strange processor might cause something like that.

destination_middleware.go
```diff
@@ -498,3 +505,239 @@ func (d *destinationWithRecordFormat) Write(ctx context.Context, recs []opencdc.
...
+	switch data := data.(type) {
+	case opencdc.StructuredData:
+		return data, nil // already decoded
+	case opencdc.RawData: // let's decode it
+	default:
+		return nil, fmt.Errorf("unexpected data type %T", data)
+	}
+
+	// check if we have the schema in the cache
+	if _, ok := d.subjectVersionCache[subject]; !ok {
```
hariso, line 724: The schema client (for standalone plugins) and the schema registry (for built-in connectors) should provide caching, or no?

destination_middleware.go
lovromazgon: I'm not sure what you mean, as far as I can see we are handling all possible cases. We tolerate when both metadata fields are empty, but we don't accept only one being empty. Am I missing something?

lovromazgon: I'm thinking now if we should skip the middleware unless both are present 🤔 Then the connector can figure out what to do with that record.

destination_middleware.go
lovromazgon: Sure in Conduit we have that cached client, but in standalone plugins we will always do a gRPC call to Conduit. Granted, it's a call on the same machine, but given that this is in the hot path (executed for every record) I'm afraid it would have a noticeable effect. I remember we got quite a nice performance boost when we went from regular gRPC calls to streams for records (a long time ago), doing a call to Conduit here would essentially undo that performance gain.

destination_middleware.go
lovromazgon: Ok I went with this: https://github.com/ConduitIO/conduit-connector-sdk/pull/155/commits/48b1d5d8d58fdc9a897c395be40438c2a432d4e8

So essentially we log the warning and skip the middleware unless the subject _and_ the version are both attached.

WDYT?

hariso: Logging sounds good to me!

destination_middleware.go
hariso: Right, that's the Conduit client. I agree, we definitely need caching.

Not for this PR, but maybe we could add caching to the sdkschema.Get()? That would benefit connectors that call the function directly too.

hariso (approved):
:zap:

destination_middleware.go
lovromazgon: Hmm, yeah, maybe we can move it. And if we move it, we can make it so it's only applied in standalone plugins, since builtin ones will already benefit from the internal cache in Conduit 👍

PR #159, opened 2024-07-30
Add default goleak option

lovromazgon:
### Description

We introduced of a [global serde cache](https://github.com/ConduitIO/conduit-commons/blob/81748fb51964f4aff6c56fb88df5bd0e9ec24c24/schema/schema.go#L100-L103) in conduit-commons, which spawns a goroutine that is taking care of cleaning the cache in the background. We need to ignore the known goroutine in acceptance test verifications of goroutine leaks.

### Quick checks:

  update/change.
- [X] I have written unit tests. (tested with existing tests in the kafka connector)


hariso (approved):
:rocket:

PR #162, opened 2024-08-01
Allow provided schema, schema cache

### LINKED ISSUE #158: [Schemas] Add caching to the schema.Get() function
### Feature description

In standalone plugins, `schema.Get()` uses a gRPC call to get a schema.  To avoid the cost of repeatedly fetching or creating the same schema many times (especially over gRPC), schemas should be cached by the SDK.

Reference: https://github.com/ConduitIO/conduit/blob/main/docs/design-documents/20240430-schema-support.md#requirements

lovromazgon:
### Description

This PR contains two additions to schemas:

- Allows the connector to return a record with structured data and an attached schema. This means the connector doesn't have to take care of encoding the data, it can only provide the schema and the middleware will use the schema to encode the data.
  - The use case are DB connectors like Postgres, which can do a better job extracting the schema directly from the DB. Those can simply attach the extracted schema to the record and return structured data as always, the middleware will take care of the encoding.
- Adds caches around the standalone schema service.
  - Note that the in-memory schema service is also cached, as it would otherwise not reflect the correct behavior, it's not smart enough to detect the same schema and will always increment the version and ID on `CreateSchema`. The cache prevents this.

Fixes #158

### Quick checks:

  update/change.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -48,6 +48,7 @@ issues:
         - errname
         - forcetypeassert
         - dupl
+        - maintidx
 
 linters:
   # please, do not use `enable-all`: it's deprecated and will be removed soon.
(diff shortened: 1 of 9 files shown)


hariso (approved):
:zap:

PR #165, opened 2024-08-02
chore: consolidate environment variables

raulb:
### Description

Follow up of https://github.com/ConduitIO/conduit-connector-protocol/pull/119

### Quick checks:

### DIFF
--- a/serve.go
+++ b/serve.go
@@ -87,12 +87,12 @@ func getPluginConfig() (pconnector.PluginConfig, error) {
 		return pconnector.PluginConfig{}, missingEnvError(pconnutils.EnvConduitConnectorToken, "v0.11.0")
 	}
 
-	connectorID := os.Getenv(pconnutils.EnvConduitConnectorID)
+	connectorID := os.Getenv(pconnector.EnvConduitConnectorID)
 	if connectorID == "" {
-		return pconnector.PluginConfig{}, missingEnvError(pconnutils.EnvConduitConnectorID, "v0.11.0")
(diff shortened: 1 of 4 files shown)


serve.go
```diff
@@ -118,7 +118,7 @@ func connectorUtilitiesGRPCTarget() (string, error) {
 // The function returns the TRACE level if the environment variable is not
 // specified or empty (that's the zerolog default).
 func connectorLogLevel() zerolog.Level {
-	level := os.Getenv(pconnutils.EnvConduitLogLevel)
+	level := os.Getenv(pconnector.EnvConduitConnectorLogLevel)
```
lovromazgon, line 121: I see that we store this env var already in `pconnector.PluginConfig`, it might make sense to use that value to parse the log level, so we have a single place where we get env vars. So I'm thinking something like this in `serve`:
```
func serve(c Connector) error {
	cfg, err := getPluginConfig()
	if err != nil {
		return fmt.Errorf("failed to get plugin configuration: %w", err)
	}
	initStandaloneModeLogger(cfg) // pass the whole cfg, let the method parse the log level as it wants
	...
}
```

lovromazgon (approved):
I added one suggestion, but it's not a blocker for this PR 👍

serve.go
raulb: @lovromazgon I took a slightly different approach https://github.com/ConduitIO/conduit-connector-sdk/pull/165/commits/75b1d9e6c6fb3c3680b48fe113168fcd89cc4351. I think this is basically what you meant with a small difference.

PR #174, opened 2024-08-12
Exported function to get connector ID

hariso:
### Description

The connector ID is useful for identifying a connector in logs, metrics, etc. in a 3rd party system. One example is the connection name in NATS JetStream.

### Quick checks:


raulb (approved):
@hariso could it maybe make more sense to actually move the internal implementation to here?

hariso:
@raulb Hmm, when I was trying it out I was convinced it would create a circular dependency, you're actually right.

raulb (approved):

internal/context.go
```diff
@@ -23,21 +23,11 @@ import (
...
-// ConnectorIDFromContext fetches the connector ID from the context. If the
-// context does not contain a connector ID it returns an empty string.
-func ConnectorIDFromContext(ctx context.Context) string {
-	connectorID := ctx.Value(connectorIDCtxKey{})
-	if connectorID != nil {
-		return connectorID.(string) //nolint:forcetypeassert // only this package can set the value, it has to be a string
-	}
-	return ""
+	return context.WithValue(ctx, ConnectorIDCtxKey{}, connectorID)
 }
```
lovromazgon, line 31: Personally, I'd prefer if we leave the struct private and simply call `internal.ConnectorIDFromContext` in the introduced function. Then we have the usage of the context key struct nicely encapsulated and both setter and getter are here in the same place. My 2 cents 🤷

hariso: i.e. like it was [here](https://github.com/ConduitIO/conduit-connector-sdk/pull/174/commits/5936fbca685ec76e5a34bd794d968a5d0f79c16d)? I don't mind either way.

lovromazgon: Yes, feels cleaner compared to exposing the context key type IMO.

hariso: Done :+1:

lovromazgon (approved):

PR #196, opened 2024-10-22
Encode key and payload whenever a schema is available

hariso:
### Description

Fixes https://github.com/ConduitIO/conduit-connector-sdk/issues/197.

Previously, encoding a record's data was tied to the flag enabling schema extraction. This PR makes sure that encoding happens whenever there's a schema available and the data is structured. Since extraction and encoding are separate actions, the middleware has also been split.

### Quick checks:

### DIFF
--- a/source_middleware.go
+++ b/source_middleware.go
@@ -669,6 +669,7 @@ func (s *sourceWithEncoding) schemaForKey(ctx context.Context, rec opencdc.Recor
 		// not both, this isn't valid.
 		return nil, fmt.Errorf("found metadata fields %v=%v and %v=%v, expected key schema subject and version to be both set to valid values, this is a bug in the connector", opencdc.MetadataKeySchemaSubject, subject, opencdc.MetadataKeySchemaVersion, version)
 	default:
+		//nolint:nilnil // we return nil to indicate that no schema is attached
 		return nil, nil
 	}
 }
@@ -739,6 +740,7 @@ func (s *sourceWithEncoding) schemaForPayload(ctx context.Context, rec opencdc.R
 		// not both, this isn't valid.
 		return nil, fmt.Errorf("found metadata fields %v=%v and %v=%v, expected payload schema subject and version to be both set to valid values, this is a bug in the connector", opencdc.MetadataPayloadSchemaSubject, subject, opencdc.MetadataPayloadSchemaVersion, version)
 	default:
+		//nolint:nilnil // we return nil to indicate that no schema is attached
 		return nil, nil
 	}
 }
--- a/source_middleware.go
+++ b/source_middleware.go
@@ -618,7 +618,7 @@ func (s *sourceWithEncoding) encodeKey(ctx context.Context, rec *opencdc.Record)
 	if _, ok := rec.Key.(opencdc.StructuredData); !ok {
 		// log warning once, to avoid spamming the logs
 		s.keyWarnOnce.Do(func() {
-			Logger(ctx).Warn().Msgf(`record key is not structured, consider disabling the source schema key encoding using "%s: false"`, configSourceSchemaExtractionKeyEnabled)
+			Logger(ctx).Warn().Msg(`record keys produced by this connector are not structured and won't be encoded"`)
 		})
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


source_middleware.go
```diff
@@ -598,6 +576,179 @@ func (s *sourceWithSchemaContext) LifecycleOnDeleted(ctx context.Context, config
...
+		// The connector has attached the schema subject and version, we can use
+		// it to retrieve the schema from the schema service.
+		return schema.Get(ctx, subject, version)
+	case subject != "" || version > 0:
+		// The connector has attached either the schema subject or version, but
+		// not both, this isn't valid.
+		return schema.Schema{}, fmt.Errorf("found metadata fields %v=%v and %v=%v, expected key schema subject and version to be both set to valid values, this is a bug in the connector", opencdc.MetadataKeySchemaSubject, subject, opencdc.MetadataKeySchemaVersion, version)
+	default:
+		// todo handle case
+		return schema.Schema{}, nil
```
lovromazgon, line 667: If there's no schema attached we should simply skip the encoding part.

source_middleware.go
```diff
@@ -598,6 +576,179 @@ func (s *sourceWithSchemaContext) LifecycleOnDeleted(ctx context.Context, config
...
+	}
+
+	return rec, nil
+}
+
+func (s *sourceWithEncoding) encodeKey(ctx context.Context, rec *opencdc.Record) error {
+	if _, ok := rec.Key.(opencdc.StructuredData); !ok {
+		// log warning once, to avoid spamming the logs
+		s.keyWarnOnce.Do(func() {
+			Logger(ctx).Warn().Msgf(`record key is not structured, consider disabling the source schema key encoding using "%s: false"`, configSourceSchemaExtractionKeyEnabled)
```
lovromazgon, line 621: Well, there's no way to disable the encoding now, it always happens 😓 I think it would be good to still have the ability to skip this, if nothing else for performance reasons when you don't need schemas. That said, I wouldn't want us to add additional parameters for that, we already have parameters for enabling schema support in the other middleware. So then the question is, do we just use those? But that seems fishy, because this middleware would be using parameters that another middleware is declaring 😕

hariso: > That said, I wouldn't want us to add additional parameters for that, we already have parameters for enabling schema support in the other middleware.

I guess you mean the middleware that extracts the schema?

But regarding disabling, yeah, this was a copy-paste from the extraction middleware. 

I agree that it would be good to make it possible to disable record data encoding. As for the params, extraction, and encoding are separate steps (and extraction might even be there), so separate params should be better IMO.

lovromazgon: Yes, I mean `sdk.schema.extract.payload.enabled` and `sdk.schema.extract.key.enabled`. My goal is to make this as simple as possible, as it's already complex, so I'm trying to see if we can avoid separate parameters.

Let's think this through. I'll only write about one parameter (`sdk.schema.extract.payload.enabled`), but the same logic applies to the key. The way it works in this PR - the value of `sdk.schema.extract.payload.enabled` is set to `true` by default, so the schema will be always extracted from the structured data. If a connector attaches its own schema (like `postgres` does), it would also have to disable this default setting, otherwise the schema would be extracted and overwritten. You're saying we should have a separate parameter for disabling only the encoding, but I'm thinking if the user ever wants to extract the schema _without_ encoding the data? I don't see when this would be the case, since we might lose precision when sending the structured data over gRPC to Conduit, it's not really a safe thing to do.

So what I'm proposing is to keep the behavior from the user's perspective and from the connector developer's perspective as it was. This means that `sdk.schema.extract.payload.enabled` is enabled by default, which means the schema gets extracted _unless_ the connector attached a schema already (extraction is a fallback). The encoding then only happens if a schema was attached, regardless if that was done by the middleware or the connector. The user can then disable the schema extraction and encoding using one switch.

Let me know if that makes sense.

hariso: > You're saying we should have a separate parameter for disabling only the encoding, but I'm thinking if the user ever wants to extract the schema without encoding the data? 

Well, I think that if we want to make it possible to disable the encoding, then I think we should have a separate parameter.

Now, should we make it possible to disable the encoding? I agree, if a user wants the schema extracted, then he probably also wants the data encoded too (unless the situation is such that there's no risk of losing precision, and the schema is needed to create a destination collection). What if a connector is adding a schema on its own, and a user doesn't want it? I'm not sure if that's something that could actually happen. All in all, if we ever face such use cases, we can think about a solution, but for now, it looks like having a schema should imply a need to encode the data.

> The encoding then only happens if a schema was attached, regardless if that was done by the middleware or the connector.

Yup, it does make sense, and it's actually how it works now (the above lines were only a copy-paste for logging purposes), so I can just leave it, and clarify the log message.

lovromazgon (approved):

lovromazgon (commented):
Actually, before merging, can you please manually try it out with the postgres connector and make sure it fixes this https://github.com/ConduitIO/conduit-connector-sdk/issues/197?

hariso:
> Actually, before merging, can you please manually try it out with the postgres connector and make sure it fixes this #197?

Good callout :+1: . Just tested Conduit with the pipeline from #197 and this fix (and no changes in the Postgres connector), and I can see the Avro data.

PR #211, opened 2024-11-15
specgen

hariso:
specgen is a tool that builds a YAML that contains all the important information about a connector. The file, called `connector.yaml`, is build from the source and destination configuration structs of a connector.

This code combines the code from https://github.com/ConduitIO/conduit-connector-sdk/pull/198 (which is the major part) plus some additional tests.

A detailed description of `specgen` can be found here: https://github.com/ConduitIO/conduit-connector-sdk/pull/198.


raulb (approved):
Approving ahead of time to unblock @hariso

PR #212, opened 2024-11-19
 specgen: re-enable acceptance tests, benchmark tests, etc.

### LINKED ISSUE #210: specgen: re-enable acceptance tests
### Feature description

The acceptance testing has been commented out in #198. We need to re-enable those tests.

hariso:
### Description

Closes #210.

### Quick checks:


acceptance_testing.go
```diff
@@ -1124,5 +1123,60 @@ func (a acceptanceTest) context(t *testing.T) context.Context {
...
+	return configureSource(
+		ctx,
+		src,
+		a.driver.SourceConfig(t),
+		a.driver.Connector().NewSpecification().SourceParams,
+	)
+}
+
+func (a acceptanceTest) configureDestination(ctx context.Context, t *testing.T, dest Destination) error {
+	return configureDestination(
```
raulb, line 1136: I feel like this either needs to be a method or a function, but why having both?

acceptance_testing.go
```diff
@@ -1124,5 +1123,60 @@ func (a acceptanceTest) context(t *testing.T) context.Context {
 	return ctx
 }
 
+func (a acceptanceTest) configureSource(ctx context.Context, t *testing.T, src Source) error {
+	return configureSource(
```
raulb, line 1127: Same as with `configureDestination`. I wonder if we could leave this either as a method or as a function.

source_middleware.go
```diff
@@ -309,12 +326,12 @@ func (s *sourceWithSchemaExtraction) extractPayloadSchema(ctx context.Context, r
 }
 
 func (s *sourceWithSchemaExtraction) schemaForType(ctx context.Context, data any, subject string) (schema.Schema, error) {
-	srd, err := schema.KnownSerdeFactories[s.config.SchemaType].SerdeForType(data)
+	srd, err := schema.KnownSerdeFactories[s.config.SchemaType()].SerdeForType(data)
```
raulb, line 329: I think we should check whether `s.config.SchemaType()` returns a valid known type.

raulb (commented):
I wonder if the second `specgen` directory is really necessary. Left other comments. Let me know what you think @hariso

acceptance_testing.go
hariso: It's just to make it a bit simpler to configure a destination that's in an `acceptanceTest`, because an `acceptanceTest` can fetch the configuration and the parameters itself.

source_middleware.go
hariso: We rely on that `s.config.SchemaType()` converts `s.config.SchemaTypeStr` to a `schema.Type`, and `SchemaTypeStr` has already been validated. But I think you have a point, I can add a test or two to make sure `s.config.SchemaType()` is doing the conversion successfully.

raulb: @hariso Just checking in. You haven't added a test for this, did you? Just in case I'm not seeing it.

raulb (approved):
Approving subject to a comment I left regarding a test for `SchemaType()`

source_middleware.go
hariso: Ah, good catch, I forgot about this comment.:) I'll add a test or two now.

Issue #54, opened 2023-01-10
Support for duration validations


lovromazgon:
### Feature description

Validations that work on floats should work on durations as well (greater than, less than).

Example:
```
type Config struct {
	MyDuration time.Duration `validate:"gt=1s,lt=5s"`
}
```

hariso:
We'll be slating this for the next release.

PR #217, opened 2024-11-27
specgen: re-enable unit tests

### LINKED ISSUE #214: specgen: re-enable unit tests, benchmark tests, etc.
### Feature description

Some of those tests have been commented out in #198. We need to re-enable those tests.

hariso:
### Description

Closes #214.

### Quick checks:


destination_middleware.go
```diff
@@ -425,26 +425,22 @@ type destinationWithSchemaExtraction struct {
...
-	keyEnabled     bool
-
 	payloadWarnOnce sync.Once
 	keyWarnOnce     sync.Once
 }
 
 func (d *destinationWithSchemaExtraction) Write(ctx context.Context, records []opencdc.Record) (int, error) {
-	if d.keyEnabled {
-		for i := range records {
+	for i := range records {
```
hariso, line 433: The change here is about using a single loop for all the records, where we decode the key and/or payload in each iteration. Previously, we used two loops: one to decode the keys, and one to decode the payloads.

destination_middleware.go
```diff
@@ -126,7 +126,7 @@ type DestinationWithBatch struct {
 	// Maximum size of batch before it gets written to the destination.
 	BatchSize int `json:"sdk.batch.size" default:"0" validate:"gt=-1"`
 	// Maximum delay before an incomplete batch is written to the destination.
-	BatchDelay time.Duration `json:"sdk.batch.delay" default:"0" validate:"gt=-1"`
+	BatchDelay time.Duration `json:"sdk.batch.delay" default:"0"`
```
raulb, line 129: Why removing the previous validation?

source_middleware.go
```diff
@@ -139,6 +139,10 @@ type SourceWithSchemaExtraction struct {
 // todo: use https://github.com/ConduitIO/conduit-commons/issues/142 to parse
 // the string value into schema.Type directly.
 func (c *SourceWithSchemaExtraction) SchemaType() schema.Type {
+	if c.SchemaTypeStr == "" {
+		return schema.TypeAvro
```
raulb, line 143: (thinking out loud) Could we initialize it with this value when instantiating `SourceWithSchemaExtraction`?

raulb (approved):
Just a question

destination_middleware.go
hariso: Because it actually isn't something we support: https://github.com/ConduitIO/conduit-connector-sdk/issues/54.

source_middleware.go
hariso: I think that would work if we constraint ourselves to always use a constructor method. It's not terrible, but unless we make the type unexported we can't really force anyone to use the constructor method. I believe that in most cases this should be enough, because when the config is parsed using the `sdk.Util.ParseConfig` function, the default should be set (because we have the `default` tag).

PR #229, opened 2024-12-24
 ParseConfig should call the custom Validate method

hariso:
### Description

In https://github.com/ConduitIO/conduit-connector-sdk/pull/198, a method for custom validation of configuration structs was added.

Given that callers of ParseConfig also want to validate the method, and that ParseConfig already does some validations (based on parameter specs), ParseConfig should call the Validate() method on a config, if it exists.

### Quick checks:


util.go
```diff
@@ -90,7 +90,19 @@ func parseConfig(
 
 	logger.Debug().Type("target", target).Msg("decoding configuration into the target object")
 	//nolint:wrapcheck // error is already wrapped by DecodeInto
```
lovromazgon, line 92: Let's move this into the `if `statement below (line 95) and not wrap the returned error.

lovromazgon (approved):
👍

util.go
hariso: You meant this: a7c327d?

lovromazgon: Yeah, that's it 👍 Thanks!

PR #246, opened 2025-01-29
Combine redmegen and specgen into conn-sdk-cli

lovromazgon:
### Description

Instead of having two tools, this introduces a single tool called `conn-sdk-cli` which we can extend in the future with more commands. The functionality of `readmegen` and `specgen` stay the same.

This PR builds on https://github.com/ConduitIO/conduit-connector-sdk/pull/115.


conn-sdk-cli/main.go
```diff
@@ -0,0 +1,72 @@
+// Copyright © 2024 Meroxa, Inc.
```
hariso, line 1: Maybe 2025 here?

hariso (approved):
LGTM, nice change!

conn-sdk-cli/main.go
lovromazgon: New laptop, I need to set up that the header is added automatically 🙈

PR #248, opened 2025-01-30
sourceWithEncoding middleware works impropertly in combination with sourceWithBatch

derElektrobesen:
### Description

sourceWithEncoding works not propertly in combination with sourceWithBatch middleware.

sourcePluginAdapter in runRead method first tries to call ReadN method and if it fails with Unimplemented error, it fallbacks on Read method.
If Source implementation supports ReadN method, this method will be called instead of Read and the result will not be encoded propertly.

If Source doesn't implement ReadN method, but supports batches using sourceWithBatch middleware the result is the same: records are not encoded.

Fixes # (issue)

### Quick checks:

### DIFF
--- a/source_middleware.go
+++ b/source_middleware.go
@@ -433,14 +433,37 @@ func (s *sourceWithEncoding) Read(ctx context.Context) (opencdc.Record, error) {
 		return rec, err
 	}
 
-	if err := s.encodeKey(ctx, &rec); err != nil {
-		return rec, err
-	}
-	if err := s.encodePayload(ctx, &rec); err != nil {
+	if err := s.encode(ctx, &rec); err != nil {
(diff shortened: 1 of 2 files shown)


lovromazgon (approved):
Nice catch! Thanks for fixing it right away and providing tests, this is great 👍 I think we have the same problem with `sourceWithSchemaExtraction`, I'll address that in a separate PR.

raulb:
@derElektrobesen Congratulations, and thank you for your first contribution! 

We'd like to hear more about how you're using Conduit, if that's alright with you. We've got this Discord (https://discord.meroxa.com/) where you can connect directly with us. We'd love to know more about your use case and how we can help you with that.

PR #254, opened 2025-01-31
Add new line before description

lovromazgon:
### Description

A description is normally a multiline markdown document. Even more - I'll argue that it will most of the time start with a chapter (e.g. `## Source` or similar). For this to work, the line can't start with a HTML comment, so we add a new line in front to push the description start into a new line.

### Quick checks:

- [X] There is no other [pull request](https://github.com/ConduitIO/conduit-connector-sdk/pulls) for the same update/change.


raulb (approved):
I'm curious about whether this will be necessary in other sections, but I think we can start with this for now and see if it's needed in the future.

lovromazgon:
Given that we don't expect other fields to contain markdown or multiple lines, I think this will be it. But yeah, let's see, we can change it if we indeed discover the need.

PR #256, opened 2025-01-31
docs for readmegen and specgen

hariso:
### Quick checks:


conn-sdk-cli/README.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+The following sections describe the most important commands in more details.
+
+## readmegen
+
+The `readmegen` command is used to generate sections in a README file that
+document the connector. `readmegen` searches for certain tags in the README file
+and replaces them with appropriate values. For example, if
+`<!-- readmegen:name -->` is found in a README, then `readmegen` will replace it
```
maha-hajja, line 29: I tested the command out, and I think mentioning the flag `--write` will be helpful here, the user will mostly need it, took me a while to find the option to overwrite

maha-hajja: will merge so we have a link for the blog post, this can be added later

maha-hajja (approved):

conn-sdk-cli/README.md
maha-hajja: @hariso

PR #258, opened 2025-02-03
readmegen: show if parameter is required or optional

hariso:
### Quick checks:


conn-sdk-cli/readmegen/testdata/test1want.md
```diff
@@ -0,0 +1,189 @@
...
+          param2: "0"
+```
+<!-- /readmegen:source.parameters.yaml -->
+
+<!-- readmegen:source.parameters.table -->
+<table class="no-margin-table">
+  <tr>
+    <th>Name</th>
+    <th>Type</th>
+    <th>Required/Optional</th>
```
raulb, line 40: Would it make sense to maybe just leave `Required` and the value would be `yes` or `blank`? What do you think? I personally think it's strange to have `Required / Optional` and only fill the value when it's required (and not optional). Having both would also make it harder to read I think.

conn-sdk-cli/readmegen/generate.go
```diff
@@ -86,10 +86,45 @@ func Generate(opts GenerateOptions) error {
...
 var funcMap = template.FuncMap{
-	"formatCommentYAML": formatCommentYAML,
-	"formatValueYAML":   formatValueYAML,
-	"zeroValueForType":  zeroValueForType,
-	"args":              args,
+	"formatCommentYAML":            formatCommentYAML,
+	"formatValueYAML":              formatValueYAML,
+	"zeroValueForType":             zeroValueForType,
+	"args":                         args,
+	"describeParameterRequirement": describeParameterRequirement,
```
raulb, line 93: Could this be something like `isRequired` and the function return a boolean?

conn-sdk-cli/readmegen/testdata/test1want.md
hariso: I wanted to have "required" and "optional" for two reasons: (1) when you're going through table and see "required" or "optional", it's pretty clear what it's about. If you see "yes" or "no", a bit less. (2) "required/optional" is a bit shorter then "required: yes" and "required: no".

lovromazgon: I agree with Raul, I feel like having both makes it harder for me to read at a glance. I'll make an even more controversial suggestion - how about we make use of some icons? Instead of this:
Screenshot 2025-02-04 at 13 24 10

We'd have something like this:
Screenshot 2025-02-04 at 13 25 42

To me the icons make required params stand out more 🤷

conn-sdk-cli/readmegen/generate.go
lovromazgon: I'm also for `isRequired`, even if it makes the template a bit more verbose (which, tbh, will make it clearer IMO). Think of readmegen as a tool that can be used in different settings, we already use it to generate documentation on conduit.io for instance. I wouldn't tailor functions to the one template that we are using right now, but rather provide general functions that can be useful in many templates.

conn-sdk-cli/readmegen/testdata/test1want.md
hariso: I guess it's a matter of preference. I like reading a row as a sentence: param1 is a string, required, etc. But I can change this into a yes/no or an icon.

conn-sdk-cli/readmegen/generate.go
hariso: We can do it that way too.:)

conn-sdk-cli/readmegen/templates/parameters.table.tmpl
```diff
@@ -20,6 +21,11 @@
 </td>
 <td>
 
+{{ if isParameterRequired $param }}✔{{ end }}
```
lovromazgon, line 24: Let's make it stand out a bit more!
```suggestion
{{ if isParameterRequired $param }}✅{{ end }}
```

conn-sdk-cli/readmegen/templates/parameters.yaml.tmpl
```diff
@@ -11,6 +11,7 @@ pipelines:
         {{- range $param := .parameters }}
           {{ formatCommentYAML $param.description 10 }}
           # Type: {{ $param.type }}
+          # Required: {{- if isParameterRequired $param }} yes{{ else }} no{{ end }}
```
lovromazgon, line 14: The blank spaces caught my eye and got me wondering why we need them, here's my suggestion to make it clearer what gets produced here (it doesn't change the output):
```suggestion
          # Required: {{ if isParameterRequired $param }}yes{{ else }}no{{ end }}
```

lovromazgon (approved):
Some nitpicks, but looks good!

conn-sdk-cli/readmegen/templates/parameters.yaml.tmpl
hariso: Ah, forgot to change this `- if` too.:)

PR #265, opened 2025-02-05
Update README.md

alarbada:
### Description

I might be just using the cli wrong, but when I ran `$ conn-sdk-cli readmegen -w` I got:

`Error: failed to generate readme: could not preprocess readme file ./README.md: readmegen close tag not found`


hariso (approved):
You're right, thanks @alarbada !

PR #267, opened 2025-02-07
disable batch delay validations on SourceWithBatch middleware

derElektrobesen:
### Description

Duration validations are not supported by conduit-common (https://github.com/ConduitIO/conduit-commons/issues/169)
This bug (?) doesn't allow to use batching.

There is no such validation in DestinationWithBatch and looks like this validation is pretty useless.

### Quick checks:


hariso (approved):
Good catch and thanks for the PR @derElektrobesen! This validation was present in DestinationWithBatch, and I removed it, but I missed this one.

PR #291, opened 2025-04-09
fix: fail on missing template keys

### LINKED ISSUE #216: Bug: unexpected output when using an unknown property key in sdk.record.format.options
### Bug description

When specifying a property that doesn't exist in the passed record, conduit writes the whole record instead of erroring.

### Steps to reproduce

Running this pipeline configuration should be enough to reproduce the error:

```yaml
version: "2.2"
pipelines:
  - id: example-pipeline
    status: running
    name: "s3-to-file"
    connectors:
      - id: example
        type: source
        plugin: generator
        settings:
          rate: 1
          format.type: structured
          format.options.id: int
          format.options.name: string
          operations: create

      - id: file
        type: destination
        plugin: "file"
        settings:
          path: ./destination.txt
          sdk.record.format: template

          # This works as expected
          sdk.record.format.options: '
            {{ printf "%v" .Payload.After }}
          '

          # This makes the pipeline fail
          sdk.record.format.options: '
            {{ printif "%v" .Payload.After }}
          '

          # But this doesn't.
          sdk.record.format.options: '
            {{ printf "%v" .Payload.Afterrrr }}
          '
          
          # And it doesn't matter whether there was previous output written or
          # not, the whole record is still written:
          sdk.record.format.options: '
            {{ printf "%v" .Key }}
            {{ printf "%v" .SomePropThatDoesNotExist }}
          '

          # Not using yaml multilines, and the issue still happens
          sdk.record.format.options: '{{ printf "%v" .SomePropThatDoesNotExist }}'

```

### Version

v0.12.2 linux/amd64

AdamHaffar:
## Description
This PR fixes #216 by adding the `missingkey=error` option to the template configuration. This ensures that template execution fails when encountering missing keys instead of silently writing the entire record.

## Fixes
- Added `missingkey=error` to template configuration
- Improved error messages for template parsing and execution
- Added comprehensive test cases for missing key scenarios

## Quick Checks
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)
- [x] I have run `make lint` and `make test` and all tests pass
- [x] I have added a changelog entry if this is a user-facing change

### DIFF
--- a/record_serializer_test.go
+++ b/record_serializer_test.go
@@ -15,6 +15,7 @@
 package sdk
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/conduitio/conduit-commons/opencdc"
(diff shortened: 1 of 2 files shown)


record_serializer_test.go
```diff
@@ -358,3 +359,68 @@ func TestTemplateRecordSerializer(t *testing.T) {
...
+			record: opencdc.Record{
+				Key: opencdc.RawData("test-key"),
+			},
+			expectError: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			serializer := TemplateRecordSerializer{}
```
lovromazgon, line 408: The assertions in this test could be simplified if we used `is`, like we do in most of our tests. Here's a [random example](https://github.com/ConduitIO/conduit-connector-sdk/blob/ca006803b8f497821949c0801660fd153cd46eb6/destination_middleware_test.go#L140-L147).

lovromazgon (commented):
Looks good, just one nitpick on the test code, to make it more consistent with our other tests. Thanks @AdamHaffar!

lovromazgon (approved):

Issue #299, opened 2025-05-09
Make user aware when sdk.batch.size is specified but sdk.batch.delay isn't


hariso:
When `sdk.batch.size` is specified but `sdk.batch.delay` isn't, the SDK will wait indefinitely for the batch size to fill up.

We should make the user aware of that (docs, and a warning /failed validation).

Tasks:

- [x] Update docs (done as part of https://github.com/ConduitIO/conduit-site/pull/298)
- [x] Log a warning https://github.com/ConduitIO/conduit-connector-sdk/pull/309
- [  ] Remove 0 as default value.

lovromazgon:
We have a batching [documentation page](https://conduit.io/docs/using/connectors/configuration-parameters/batching), but this behavior is not called out.

Making it required to specify both could be a good way of preventing this pitfall, although we should still allow the user to set the delay to `0s` if they want to. In a sufficiently fast stream, the delay would just add unnecessary overhead, so skipping it might be preferable in that case. That said, we can at least print a warning when the connector starts, if we notice such a configuration.

So my proposal:
- Document it on the [documentation page](https://conduit.io/docs/using/connectors/configuration-parameters/batching) (make it stand out, a warning admonition).
- Output a warning in the [middleware](https://github.com/ConduitIO/conduit-connector-sdk/blob/5e067ac3c3e25e2c3fdab129dbedda894f020d59/source_middleware.go#L693) in `Open` if only one of the batch parameters is greater than 0.
- Make `sdk.batch.size` and `sdk.batch.delay` required together (either the user sets both or none). Note that the connector needs to call the `Validate` method properly if they override it (related to https://github.com/ConduitIO/conduit-connector-sdk/issues/264).

hariso:
@lovromazgon Could you elaborate a bit on this:

> Making it required to specify both could be a good way of preventing this pitfall, although we should still allow the user to set the delay to `0s` if they want to. In a sufficiently fast stream, the delay would just add unnecessary overhead, so skipping it might be preferable in that case.

and

>     * Make `sdk.batch.size` and `sdk.batch.delay` required together (either the user sets both or none). 

It looks like you're suggesting that we detect if a user explicitly sets the delay to 0s (0s is also the default). However, it doesn't seem like that's possible, because the SDK is setting the default values into a configuration struct, so we can't really tell who set a value.

As for setting a size without a delay for a fast stream, I'd still say it's good to have a batch delay, because any stream can stop or become slow for whatever reason.

[...]

PR #309, opened 2025-05-28
Warn if one of sdk.batch.size or sdk.batch.delay is set to 0

hariso:
### Description

Part of https://github.com/ConduitIO/conduit-connector-sdk/issues/299.

### Quick checks:

### DIFF
--- a/source_middleware.go
+++ b/source_middleware.go
@@ -696,7 +696,7 @@ func (s *sourceWithBatch) Open(ctx context.Context, pos opencdc.Position) error 
 	}
 	if *s.config.BatchSize == 0 && *s.config.BatchDelay > 0 {
 		Logger(ctx).Warn().
-			Msg("batch delay is set but batch size is not, this might result in extremely large batches for fast sources")
+			Msg("sdk.batch.delay is set but sdk.batch.size is not, this might result in extremely large batches for fast sources")
 	}
 
 	// Batching configured, which means that we need to collect records from the source


hariso:
@nickchomey You're making a good point... I've written the code like this to be consistent with what we have here:

```go
	// Batching configured, which means that we need to collect records from the source
	// until the batch is full or the batch delay is reached.
	if *s.config.BatchSize > 0 || *s.config.BatchDelay > 0 {
		s.collectFn = s.collectWithReadN
		go s.runReadN(ctx)
	} else {
		// Batching not configured, simply forward the call.
		s.collectFn = s.Source.ReadN
	}
```

From what I could understand (I'll need to confirm), a zero batch size is used to indicate that batching is disabled, hence the code above. Unless I'm missing something, I agree with you, 1 should mean no batching.

I'll be revisiting this code anyway to add some more validation, so I can take a look at better defaults.

hariso:
> Batchsize of 1 is "dont batch, send them individually". 2+ is batching. This PR should probably be such that 0 and 1 are handled and warned/errored as such.

:100: Also, I agree, batch size 0 shouldn't be allowed. This PR is meant to warn users if one value has been set and the other hasn't, since it might result in unexpected behavior. As for the defaults/disabling, I can create an issue so we can find something that works for most. Thanks for highlighting this!

> So, I dont see an issue, really, with setting the delay to 0.

Yup, as long as a batch size > 0, it doesn't really matter. But, if the delay is set, but the batch size isn't, then a fast producing source could build a batch that is very large.

source_middleware.go
```diff
@@ -690,6 +690,17 @@ func (s *sourceWithBatch) Open(ctx context.Context, pos opencdc.Position) error
 	s.readCh = make(chan readResponse, *s.config.BatchSize)
 	s.readNCh = make(chan readNResponse, 1)
 
+	if *s.config.BatchSize > 0 && *s.config.BatchDelay == 0 {
+		Logger(ctx).Warn().
+			Msg("batch size is set but batch delay is not, this might result in the connector waiting indefinitely for the batch to be filled")
```
raulb, line 695: I would use the actual configuration options to make it explicit

```suggestion
			Msg("sdk.batch.size is set but sdk.batch.delay is not, this might result in the connector waiting indefinitely for the batch to be filled")
```

raulb (changes requested):
@hariso small suggestion

raulb (approved):

